PR #25, opened 2017-09-21
ROC-2046: using @Money and BigDecimal instead of double field to stor…

yarpo:
https://tools.hmcts.net/jira/browse/ROC-2046

### Change description ###
BigDecimal instead of double
@Money annotation for CountyCountJudgment.paidAmount field

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/models/ccj/RepaymentPlan.java
```diff
@@ -17,15 +22,32 @@
         EVERY_MONTH
     }
 
-    private final double remainingAmount;
-    private final double firstPayment;
-    private final double instalmentAmount;
+    @Valid
```
rlewan: `@Valid` is needed only if you want to validate nested properties of the object, it can be removed. Please apply to other places you've modified in this PR as well.

src/main/java/uk/gov/hmcts/cmc/claimstore/models/CountyCourtJudgment.java
```diff
@@ -23,8 +24,9 @@
     @NotNull
     private final TheirDetails defendant;
 
-    @Min(0)
-    private double paidAmount;
+    @Valid
+    @Money
+    private BigDecimal paidAmount;
```
timja-kainos: I think this should have a minimum on it and `@Money` shouldn't have the min constraint on it

src/test/java/uk/gov/hmcts/cmc/claimstore/models/sampledata/SampleCountyCourtJudgment.java
```diff
@@ -6,12 +6,13 @@
...
 import uk.gov.hmcts.cmc.claimstore.models.otherparty.TheirDetails;
 
+import java.math.BigDecimal;
 import java.time.LocalDate;
 
 public class SampleCountyCourtJudgment {
 
     private TheirDetails defendant = SampleTheirDetails.builder().individualDetails();
-    private double paidAmount = 0.0;
+    private BigDecimal paidAmount = BigDecimal.valueOf(0);
```
rlewan: There is a `BigDecimal.ZERO` constant, would be nice to make use of it here.

src/main/java/uk/gov/hmcts/cmc/claimstore/models/CountyCourtJudgment.java
rlewan: Hi Tim I've talked with Patryk about that and at the moment all places where we use money require the value to be at least `0.01`. I agree that these two should be separate but not sure if it's worth it if it would be done for all of them.

src/test/java/uk/gov/hmcts/cmc/claimstore/constraints/DateNotInThePastValidatorTest.java
```diff
@@ -0,0 +1,35 @@
...
+@RunWith(MockitoJUnitRunner.class)
+public class DateNotInThePastValidatorTest {
+
+    @Mock
+    private ConstraintValidatorContext context;
+
+    private DateNotInThePastConstraintValidator validator = new DateNotInThePastConstraintValidator();
+
+    @Test
+    public void shouldReturnTrueForValidDateInTheFuture() {
```
timja-kainos: @pragnesh-patel can you check you're happy with the cases covered here

src/test/java/uk/gov/hmcts/cmc/claimstore/models/sampledata/SampleCountyCourtJudgment.java
yarpo: ok

src/test/java/uk/gov/hmcts/cmc/claimstore/constraints/DateNotInThePastValidatorTest.java
yarpo: The same cases are written for DateNotInTheFuture. If this is not enough - both should be updated

src/main/java/uk/gov/hmcts/cmc/claimstore/models/ccj/RepaymentPlan.java
yarpo: done

PR #32, opened 2017-09-25
ROC-2266-refactor-pdf-template content change

robertparkinson:

https://tools.hmcts.net/jira/browse/ROC-2266

### Change description ###

```
[ ] Yes
[X] No
```


src/main/resources/staff/templates/document/legalSealedClaim.html
```diff
@@ -204,7 +204,7 @@
   {% if externalReferenceNumber is defined %}
     <table class="border-bottom">
       <tr>
-        <td class="two-per-row"><h4>Reference</h4></td>
+        <td class="two-per-row"><h4>Claimant solicitor's reference</h4></td>
```
yarpo: Shouldn't we use 

```’```

instead of

```
'
```

?

timja-kainos: yes, this was confirmed on the cross gov slack, T1 only uses `’` and T2 will be moving over

PR #37, opened 2017-09-27
ROC-2272: Add feature toggle for water marking legal pdf

dharmendrak:


src/main/java/uk/gov/hmcts/cmc/claimstore/config/properties/emails/StaffEmailTemplates.java
```diff
@@ -42,6 +42,10 @@ public String getCCJRequestSubmittedEmailSubject() {
         return readBytes("/staff/templates/document/legalSealedClaim.html");
     }
 
+    public byte[] getWaterMarkedLegalSealedClaim() {
+        return readBytes("/staff/templates/document/waterMarkedLegalSealedClaim.html");
```
robertparkinson: Is it not possible to move this check to inside the html file so that the water mark only gets added if on demo? other wise we will have to maintain two template files when making changes

src/main/resources/staff/templates/document/waterMarkedLegalSealedClaim.html
```diff
@@ -0,0 +1,547 @@
+<html>
+<head>
```
timja-kainos: NO, this is not the way to do it.

Add a variable to the content provider and then check for the variable

src/integrationTest/resources/environment.properties
```diff
@@ -10,3 +10,5 @@ notifications.templates.email.defendantResponseIssued = defendant-response=issue
 notifications.templates.email.claimantResponseIssued = claimant-response-issued-template
 notifications.templates.email.staffMoreTimeRequested = staff-more-time-requested-template
 notifications.templates.email.defendantMoreTimeRequested = defendant-more-time-requested-template
+
+claim-store.watermark-pdf.enabled: true
```
timja-kainos: why do you need to enable it

src/test/java/uk/gov/hmcts/cmc/claimstore/documents/content/LegalSealedClaimContentProviderTest.java
```diff
@@ -35,4 +35,20 @@ public void shouldCreateContent() throws Exception {
         assertThat(contents).isNotEmpty().containsKey("feePaid").containsValue("£500.01");
     }
 
+    @Test
+    public void contentShouldIncludeWaterMarkFlag() throws Exception {
```
timja-kainos: this test doesn't look like it has a lot of value, you are hardcoding the value to be true, so of course its there

timja-kainos: also not sure you you aren't using the default fee amount / default SampleClaim

src/integrationTest/resources/environment.properties
timja-kainos: this shouldn't be here

timja-kainos: ???

src/main/resources/application.yml
```diff
@@ -67,6 +67,9 @@ staff-notifications:
 pdfService:
   baseUrl: ${PDF_SERVICE_URL:}
 
+feature_toggles:
+  watermark-pdf:
+    enabled: true
```
timja-kainos: NO, default to off

dharmendrak: was testing

src/test/java/uk/gov/hmcts/cmc/claimstore/documents/content/LegalSealedClaimContentProviderTest.java
dharmendrak: used

src/integrationTest/resources/environment.properties
dharmendrak: removed

src/main/java/uk/gov/hmcts/cmc/claimstore/documents/content/LegalSealedClaimContentProvider.java
```diff
@@ -17,10 +18,13 @@
 public class LegalSealedClaimContentProvider {
 
     private final StatementOfValueProvider statementOfValueProvider;
+    private final boolean watermarkPdfEnabled;
```
timja-kainos: watermarkPdf

src/main/resources/staff/templates/document/legalSealedClaim.html
```diff
@@ -28,6 +28,15 @@
       font-family: "Open Sans"
     }
 
+    {% if waterMark %}
+    body {
```
timja-kainos: maybe indent this?

PR #40, opened 2017-09-29
Fail fast when another claim with the same external ID is being saved

sabahirfan:
https://tools.hmcts.net/jira/browse/ROC-2438

### Change description ###
Making API fail fast when another claim with the same external ID is being saved

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/cmc/claimstore/services/ClaimServiceTest.java
```diff
@@ -164,6 +165,15 @@ public void saveClaimShouldFinishSuccessfully() {
         verify(eventProducer, once()).createClaimIssuedEvent(eq(createdClaim), eq(null), anyString());
     }
 
+    @Test(expected = ConflictException.class)
+    public void saveClaimShouldThrowConflictExceptionForDuplicateClaim() {
```
geek-soft: I would have preferred to have integration test for it so that we:
- test behaviour the same way as client triggers it
- test plays role of the API documentation
- when DB is engaged for integration-tests later we are additionally ensure DB transaction is not even started

sabahirfan: Done.

PR #46, opened 2017-10-03
Update deps

timja-kainos:


build.gradle
```diff
@@ -128,7 +128,7 @@ dependencies {
     testCompile group: 'org.springframework.boot', name: 'spring-boot-starter-test', version: versions.springBoot
 
     // https://mvnrepository.com/artifact/org.mockito/mockito-core
-    testCompile group: 'org.mockito', name: 'mockito-core', version: '2.9.0'
+    testCompile group: 'org.mockito', name: 'mockito-core', version: '2.10.0'
```
rlewan: I'm a little surprised we need to import this ourselves, I thought Spring testing above would import all such stuff?

timja-kainos: It does but its quite an old version, this mockito has some improvements that are good for us.

rlewan: 👍

PR #49, opened 2017-10-06
ROC-2485: blocked API from accepting response when CCJ was requested

yarpo:
https://tools.hmcts.net/jira/browse/ROC-2485

### Change description ###
API blocked for saving response when response was already submitted for a claim or CCJ was requested.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/services/DefendantResponseService.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/services/DefendantResponseService.java
@@ -44,9 +44,16 @@ public class DefendantResponseService {
             throw new ForbiddenActionException("Response for the claim " + claimId + " was already submitted");
         }
 
+        if (isCCJAlreadyRequested(claim)) {
+            throw new ForbiddenActionException(
+                "County Court Judgment for the claim " + claimId + " was already requested"
+            );
+        }
+
         final String defendantEmail = userService.getUserDetails(authorization).getEmail();
-        defendantResponseRepository.save(claimId, defendantId, defendantEmail,
-            jsonMapper.toJson(responseData));
+        defendantResponseRepository.save(
+            claimId, defendantId, defendantEmail, jsonMapper.toJson(responseData)
+        );
 
         final Claim claimAfterSavingResponse = claimService.getClaimById(claimId);
 
@@ -59,4 +66,7 @@ public class DefendantResponseService {
         return null != claim.getRespondedAt();
     }
 
+    private boolean isCCJAlreadyRequested(final Claim claim) {
+        return null != claim.getCountyCourtJudgmentRequestedAt();
+    }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/cmc/claimstore/services/DefendantResponseService.java
```diff
@@ -59,4 +66,7 @@ private boolean isResponseAlreadySubmitted(final Claim claim) {
         return null != claim.getRespondedAt();
     }
 
+    private boolean isCCJAlreadyRequested(final Claim claim) {
+        return null != claim.getCountyCourtJudgmentRequestedAt();
```
timja-kainos: this reads wrong to me,
I would normally read this as:
```
return claim.getCountyCourtJudgmentRequestedAt() != null;
```

src/test/java/uk/gov/hmcts/cmc/claimstore/services/DefendantResponseServiceTest.java
```diff
@@ -68,11 +70,28 @@ public void saveShouldFinishSuccessfully() {
...
-        responseService.save(CLAIM_ID, DEFENDANT_ID, app, AUTHORISATION);
+        responseService.save(CLAIM_ID, DEFENDANT_ID, VALID_APP, AUTHORISATION);
 
         //then
         verify(eventProducer, once())
-            .createDefendantResponseEvent(eq(claim) );
+            .createDefendantResponseEvent(eq(claim));
+    }
+
+    @Test(expected = ForbiddenActionException.class)
```
geek-soft: Would be better to subclass from `ForbiddenActionException` so then we can check specific exception. If method throws the same exception for many reasons that kind of test does not quite prove the right exception is thrown. Alternatively asserting error message would be some sort of solution.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/DefendantResponseService.java
```diff
@@ -44,9 +44,16 @@ public Claim save(
             throw new ForbiddenActionException("Response for the claim " + claimId + " was already submitted");
         }
 
+        if (isCCJAlreadyRequested(claim)) {
+            throw new ForbiddenActionException(
```
geek-soft: We could have more specific  `CCJAlreadyRequestedException ` exception subclassing from `ForbiddedActionException`.

src/test/java/uk/gov/hmcts/cmc/claimstore/services/DefendantResponseServiceTest.java
yarpo: ok

src/main/java/uk/gov/hmcts/cmc/claimstore/services/DefendantResponseService.java
yarpo: ok

PR #66, opened 2017-10-17
ROC-2513: search claim by external reference number (legal specific)

sabahirfan:
https://tools.hmcts.net/jira/browse/ROC-2513

### Change description ###
This PR adds endpoint to search claim by reference number or by externalReferenceNumber.

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/GetClaimByClaimReferenceTest.java
```diff
@@ -0,0 +1,64 @@
...
+        webClient
+            .perform(get("/claims/" + REFERENCE_NUMBER))
+            .andExpect(status().isNotFound())
+            .andReturn();
+    }
+
+    @Test
+    public void shouldReturn404HttpStatusWhenWrongUrlGiven() throws Exception {
+
+        webClient
```
timja-kainos: this test isn't specific to the controlled you are adding, its a test that would never fail unless someone adds an endpoint that matches it.
could you remove please?

src/main/resources/db/migration/V2017_10_16_1600__Add_external_reference_to_claim.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE claim
+  ADD COLUMN external_reference TEXT;
```
timja-kainos: I'm not sure if this belongs here,
we usually keep all user entered data on the claim.
You can query on json fine..

Completely up for discussion though 👍

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/GetClaimByExternalReferenceTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+        webClient
+            .perform(get("/claims/representative/" + EXTERNAL_REFERENCE))
+            .andExpect(status().isOk())
+            .andReturn();
+    }
+
+    @Test
+    public void shouldReturn500HttpStatusWhenInternalErrorOccurs() throws Exception {
+        given(claimRepository.getByExternalReference(eq(EXTERNAL_REFERENCE))).willThrow(new RuntimeException("error"));
```
yarpo: please either leave an empty line always or always do not :) #inconsistent.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/ClaimController.java
```diff
@@ -57,6 +60,19 @@ public Claim getByExternalId(@PathVariable("externalId") final String externalId
...
+    @GetMapping("/{claimReference:" + CLAIM_REFERENCE_PATTERN + "}")
+    @ApiOperation("Fetch claim for given claim reference")
+    public Claim getByClaimReference(@PathVariable("claimReference") final String claimReference) {
+        return claimService.getClaimByReference(claimReference)
+            .orElseThrow(() -> new NotFoundException("Claim not found by claim reference " + claimReference));
+    }
+
+    @GetMapping("/representative/{externalReference}")
+    @ApiOperation("Fetch user claims for given external reference number")
+    public List<Claim> getClaimByExternalReference(@PathVariable("externalReference") final String externalReference) {
```
yarpo: is `externalReference` the same as `externalId`?

src/main/resources/db/migration/V2017_10_16_1600__Add_external_reference_to_claim.sql
yarpo: as `claim` table should be in fact `case` and `external_reference` is related to everything for this case IMHO separate column is a good idea.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/ClaimController.java
yarpo: ok, I can see it is. Do we want to rename all `externalId` to `externalReference` now?

timja-kainos: no, externalID and external reference are different things.
external reference is a user entered ID by the solicitor

src/main/resources/db/migration/V2017_10_16_1600__Add_external_reference_to_claim.sql
sabahirfan: I thought of implementing search on json objects, but my concern was the performance for search query on json. Had discussion with @dharmendrak  as well yesterday. I am happy to change to json search if that has better performance. Open for better options here.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/GetClaimByClaimReferenceTest.java
sabahirfan: Agree Done.

src/main/resources/db/migration/V2017_10_16_1600__Add_external_reference_to_claim.sql
timja-kainos: I would say performance is not a concern here...

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/GetClaimByExternalReferenceTest.java
sabahirfan: done

src/main/resources/db/migration/V2017_10_16_1600__Add_external_reference_to_claim.sql
dharmendrak: separate column is good idea, as  all our searches are happening on claim elements not on claim data (json blob). we can write a index on external reference number and reference number to increase the performance of database.  ExternalId is also part of claim data, however it is extracted and stored separately. I support for separate column

timja-kainos: we can put an index on the json data too.
Performance is not a concern here, as T2 will be on CCD before that matters and then we need to solve search properly.
I still don't think it should go on its own column as it violates our current principal of user entered data on the claim data

singh-surjeet: I agree with Tim that once we're on CCD we require to redo searches. Also external-id doesn't uniquely identify a case like reference because user can enter duplicate values. I suggest having an index on jsonB column on external-Id and hold onto our structure as is.

sabahirfan: @singh-surjeet implemented search on json claim data.

timja-kainos:
fix conflicts please

src/main/java/uk/gov/hmcts/cmc/claimstore/models/Claim.java
```diff
@@ -51,7 +52,8 @@ public Claim(
         final ResponseData response,
         final String defendantEmail,
         final CountyCourtJudgment countyCourtJudgment,
-        final LocalDateTime countyCourtJudgmentRequestedAt
+        final LocalDateTime countyCourtJudgmentRequestedAt,
+        final String externalReferenceNumber
```
timja-kainos: this shouldn't be here anymore?

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/ClaimRepository.java
```diff
@@ -18,10 +18,10 @@
 public interface ClaimRepository {
 
     @SuppressWarnings("squid:S1214") // Pointless to create class for this
-    String SELECT_FROM_STATEMENT = "SELECT * FROM claim";
+        String SELECT_FROM_STATEMENT = "SELECT * FROM claim";
```
timja-kainos: indenting changes here don't look right

sabahirfan: fixed.

src/main/java/uk/gov/hmcts/cmc/claimstore/models/Claim.java
sabahirfan: Removed already

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/mapping/ClaimMapper.java
```diff
@@ -48,11 +48,11 @@ private ClaimData toClaimData(final String input) {
     }
 
     private ResponseData toNullableResponseData(final String input) {
-        return toNullableEntity(input, ResponseData.class);
```
yarpo: why toNullableEntity was wrong?

yarpo: @timja-kainos It looks like your code in git blame?

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/ClaimController.java
```diff
@@ -57,6 +60,24 @@ public Claim getByExternalId(@PathVariable("externalId") final String externalId
...
+    @GetMapping("/{claimReference:" + CLAIM_REFERENCE_PATTERN + "}")
+    @ApiOperation("Fetch claim for given claim reference")
+    public Claim getByClaimReference(@PathVariable("claimReference") final String claimReference,
+                                     @RequestHeader(HttpHeaders.AUTHORIZATION) final String authorisation) {
+
+        return claimService.getClaimByReference(claimReference, authorisation)
+            .orElseThrow(() -> new NotFoundException("Claim not found by claim reference " + claimReference));
+    }
+
+    @GetMapping("/representative/{externalReference}")
```
dharmendrak: api containing representative makes it very specific

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/mapping/ClaimMapper.java
```diff
@@ -48,11 +48,11 @@ private ClaimData toClaimData(final String input) {
...
+        return input != null ? jsonMapper.fromJson(input, ResponseData.class) : null;
     }
 
     private CountyCourtJudgment toNullableCountyCourtJudgment(final String input) {
-        return toNullableEntity(input, CountyCourtJudgment.class);
+        return input != null ? jsonMapper.fromJson(input, CountyCourtJudgment.class) : null;
```
dharmendrak: I don't know, however I would have preferred
Optional.ofNullable(input).map(I-> jsonMapper.fromJson(i, CountyCourtJudgment.class)).orElse(null)

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -70,11 +70,22 @@ public Claim getClaimByExternalId(final String externalId) {
             .orElseThrow(() -> new NotFoundException("Claim not found by external id " + externalId));
     }
 
+    public Optional<Claim> getClaimByReference(final String reference, final String authorisation) {
+        final String submitterId = userService.getUserDetails(authorisation).getId();
+        return claimRepository
+            .getByClaimReferenceAndSubmitter(reference, submitterId);
+    }
+
     public Optional<Claim> getClaimByReference(final String reference) {
```
dharmendrak: do we still need this method?? I agree we need to update T1 frontend?

sabahirfan: Yes, it is used by `isDefendantLinked`

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/mapping/ClaimMapper.java
sabahirfan: reverted as it appearing after resolving conflicts.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/mapping/ClaimMapper.java
sabahirfan: reverted as it appearing after resolving conflicts.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -70,11 +70,22 @@ public Claim getClaimByExternalId(final String externalId) {
             .orElseThrow(() -> new NotFoundException("Claim not found by external id " + externalId));
     }
 
+    public Optional<Claim> getClaimByReference(final String reference, final String authorisation) {
```
timja-kainos: we already have a method that does exactly this, remove please

PR #77, opened 2017-10-28
Run integration tests in shared DB service

geek-soft:

None

### Change description ###

This is to instantiate one database instance and run all the tests against that single instance. To avoid problems between tests runs caused by some database state tables are purged per test class.

Implementation is not ready for running tests in parallel however increasing threads executing tests right now does not bring any benefits due to the fact that by default spring tests are not ready for parallel run (that got fixed in Spring 5)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/resources/environment.properties
+++ b/src/integrationTest/resources/environment.properties
@@ -1,3 +1,6 @@
+database.driverClassName=org.testcontainers.jdbc.ContainerDatabaseDriver
+database.url=jdbc:tc:postgresql:9.6://localhost/claimstore
+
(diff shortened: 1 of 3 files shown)


src/integrationTest/resources/environment.properties
```diff
@@ -1,3 +1,6 @@
+database.driverClassName=org.testcontainers.jdbc.ContainerDatabaseDriver
+database.url=jdbc:tc:postgresql:9.6://localhost/claimstore
```
geek-soft: From my experience I noticed that database name is ignored by `org.testcontainers.jdbc.ContainerDatabaseDriver `and all tests run against database of name `test`.

PR #85, opened 2017-11-08
Use Alpine images in Dockerfiles and integration tests

geek-soft:

None

### Change description ###

Use Alpine images in Dockerfiles and integration tests. That will speedup build process by tiny bit in case local cache is empty.

```
[ ] Yes
[x] No
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8-jre
+FROM openjdk:8-jre-alpine
 
 COPY build/install/claim-store /opt/app/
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM openjdk:8-jre
+FROM openjdk:8-jre-alpine
+
+RUN apk add --no-cache curl
```
timja-kainos: wget is in alpine by default, perhaps change to using that?

PR #90, opened 2017-11-13
ROC-2825: Make free mediation property of DefendantResponse nullable

yarpo:
https://tools.hmcts.net/jira/browse/ROC-2825

### Change description ###
Free mediation not always is mandatory

```
[x] Yes
[ ] No
```


cmc-domain-model/src/main/java/uk/gov/hmcts/cmc/claimstore/models/ResponseData.java
```diff
@@ -52,7 +52,6 @@
     @Size(max = 99000)
     private final String defence;
 
-    @NotNull
     @JsonUnwrapped
     private final FreeMediationOption freeMediation;
```
geek-soft: Because FreeMediation is nullable we need to make getter return `Optional< FreeMediationOption >` so we avoid confusion later one when somebody tries to use it.

yarpo: yep, I'm working on it. It breaks quite a lot of tests. In about 10 minutes I should push my code

src/main/resources/staff/templates/email/defendantResponse/body.txt
```diff
@@ -6,7 +6,7 @@ The defendant has submitted an already paid defence which is attached as a PDF.
 You need to ask the claimant if they want to proceed with the claim.
 {% endif %}
 
-{% if response.freeMediation == 'YES' %}
+{% if response.freeMediation.orElse('NO')  == 'YES' %}
```
timja-kainos: I don't think no is right here, the text should probably not be shown

geek-soft: I am not sure this is right. For the part of the response journey where defendant is not able to choose we should not say `The defendant has chosen not to use the free mediation service.`. In this case they dod not make any choice. Better check with business.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/notifications/DefendantResponseNotificationService.java
```diff
@@ -130,7 +130,9 @@ public void logNotificationFailure(
     }
 
     private Map<String, String> aggregateParams(final Claim claim, final ResponseData responseData) {
-        boolean isFreeMediationRequested = responseData.getFreeMediation().equals(ResponseData.FreeMediationOption.YES);
+        boolean isFreeMediationRequested = responseData.getFreeMediation()
```
geek-soft: What happens when `isFreeMediationRequested` to NO? I wonder whether we should be cautious as in the scenario below.

src/main/resources/staff/templates/email/defendantResponse/body.txt
timja-kainos: isn't this what I said? 👅

geek-soft: It is however at a time of writing I wasn't able to see your comment 👅

src/main/java/uk/gov/hmcts/cmc/claimstore/services/notifications/DefendantResponseNotificationService.java
```diff
@@ -130,7 +130,14 @@ public void logNotificationFailure(
     }
 
     private Map<String, String> aggregateParams(final Claim claim, final ResponseData responseData) {
-        boolean isFreeMediationRequested = responseData.getFreeMediation().equals(ResponseData.FreeMediationOption.YES);
+        boolean isFreeMediationApplicable = responseData.getFreeMediation().isPresent();
+        boolean isFreeMediationRequested = false;
+
+        if (isFreeMediationApplicable) {
+            isFreeMediationRequested = responseData
```
jenkins-reform-hmcts: MAJOR Call "Optional#isPresent()" before accessing the value. [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AS3655)

PR #104, opened 2017-11-27
ROC-2903: Mapper from CCD case to CMC claim

sabahirfan:
https://tools.hmcts.net/jira/browse/ROC-2903

### Change description ###
This PR adds mapping of CCD case to CMC claim.

```
[ ] Yes
[x] No
```

### DIFF
--- a/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CCDCase.java
+++ b/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CCDCase.java
@@ -7,6 +7,7 @@ import lombok.Value;
 @Builder
 public class CCDCase {
 
+    private Long id;
     private String referenceNumber;
     private String submitterId;
     private String submittedOn;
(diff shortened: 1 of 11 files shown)


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CCDCase.java
```diff
@@ -7,6 +7,7 @@
 @Builder
 public class CCDCase {
 
+    private Long id;
```
rkondratowicz: is there are reason for this to not be a string?

sabahirfan: any reason to make this String?

rkondratowicz: swagger spec for ccd says it's a string.

rkondratowicz: actually, that's fine. nevermind.

PR #115, opened 2017-12-01
ROC-2683: Remove claimant type for legal

robertparkinson:
https://tools.hmcts.net/jira/browse/ROC-2683

### Change description ###
Use a single type for claimant

```
[ ] Yes
[X] No
```


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/party/Claimant.java
```diff
@@ -0,0 +1,41 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import uk.gov.hmcts.cmc.domain.models.Address;
+import uk.gov.hmcts.cmc.domain.models.legalrep.Representative;
+
+import java.util.Objects;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Claimant extends Party {
+
+    public Claimant(
```
timja-kainos: all these fields are on party?

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/party/Claimant.java
```diff
@@ -0,0 +1,41 @@
+package uk.gov.hmcts.cmc.domain.models.party;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import uk.gov.hmcts.cmc.domain.models.Address;
+import uk.gov.hmcts.cmc.domain.models.legalrep.Representative;
+
+import java.util.Objects;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Claimant extends Party {
```
timja-kainos: the legal tests will need updating too

geek-soft: If party itself is enough without identifying whether it is individual, organisation or others subtypes we should make Party non abstract and allow instantiating it. Then if someone has to use more specialised type they can select one of 4 subtypes.

Other thing is the name, having claimant subtype deriving from party where other subtypes are split using different perspective does not make sense. If the first approach is not desirable by the team then we should at least call this class `UnspecifiedParty` or similar.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/party/Party.java
```diff
@@ -29,7 +29,8 @@
         @JsonSubTypes.Type(value = Individual.class, name = "individual"),
         @JsonSubTypes.Type(value = SoleTrader.class, name = "soleTrader"),
         @JsonSubTypes.Type(value = Company.class, name = "company"),
-        @JsonSubTypes.Type(value = Organisation.class, name = "organisation")
+        @JsonSubTypes.Type(value = Organisation.class, name = "organisation"),
+        @JsonSubTypes.Type(value = Claimant.class, name = "claimant")
```
geek-soft: When other change is made type indicator should be something like `party` or `unspecified`

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/party/Claimant.java
```diff
@@ -0,0 +1,41 @@
...
+        final Address address,
+        final Address correspondenceAddress,
+        final String mobilePhone,
+        final Representative representative
+    ) {
+        super(name, address, correspondenceAddress, mobilePhone, representative);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
geek-soft: Do not override equals method. The base one is perfectly fine.

geek-soft:
Closing for now as this is no longer priority. Will get reopened after public-beta go live or when priorities change.

PR #116, opened 2017-12-01
CNP changes

timja-kainos:


build.gradle
```diff
@@ -177,7 +177,12 @@ jacocoTestReport {
 }
 
 project.tasks['sonarqube'].dependsOn test, apiTest
-project.tasks['check'].dependsOn apiTest
+check.dependsOn apiTest
+checkstyleMain.shouldRunAfter(compileJava)
+test.shouldRunAfter(checkstyleTest)
+apiTest.shouldRunAfter(checkstyleIntegrationTest)
```
geek-soft: Does it have any significant impact on time needed to run api test? The same for running checkstyle on test source set before executing unit tests?

timja-kainos: don't think so, the main problem it had was `apiTest` was always running before `test` which really annoyed me

PR #117, opened 2017-12-04
ROC-2760: Save represented claims into CCD

sabahirfan:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/ROC-2760

### Change description ###
This PR adds feign client for CCD integration.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #119, opened 2017-12-06
ROC-2665: Consume service authorisation access client

dharmendrak:


build.gradle
```diff
@@ -123,6 +123,7 @@ dependencies {
 
   compile group: 'io.github.openfeign', name: 'feign-httpclient', version: '9.5.0'
   compile group: 'uk.gov.hmcts.reform', name: 'document-management-client', version: '1.0.2'
+  compile group: 'uk.gov.hmcts.reform', name: 'service-authorisation-access-client', version: '0.0.2'
```
geek-soft: Why is this version 0.0.2. It does not give me confidence. When `service-authorisation-access-client` was developed it should have got much higher version.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -95,6 +95,7 @@ public Claim getClaimByExternalId(final String externalId) {
     @Transactional
     public Claim saveClaim(final String submitterId, final ClaimData claimData, final String authorisation) {
         final String externalId = claimData.getExternalId().toString();
+        final String authToken = userService.generateServiceAuthToken();
```
geek-soft: Where is this being used?

src/main/java/uk/gov/hmcts/cmc/claimstore/services/UserService.java
```diff
@@ -10,9 +14,13 @@
 public class UserService {
 
     private IdamApi idamApi;
+    private final AuthTokenGenerator cachedServiceAuthTokenGenerator;
```
geek-soft: Rename, it does not matter whether this is cached or not in the service context.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/UserService.java
```diff
@@ -10,9 +14,13 @@
...
 
-    public UserService(final IdamApi idamApi) {
+    @Autowired
+    public UserService(final IdamApi idamApi,
+                       final AuthTokenGenerator cachedServiceAuthTokenGenerator) {
```
geek-soft: Are you sure cached instance will get injected here?

src/main/java/uk/gov/hmcts/cmc/claimstore/services/UserService.java
```diff
@@ -25,4 +33,8 @@ public GeneratePinResponse generatePin(final String name, final String authorisa
             authorisation.replace("Bearer: ", "")
         );
     }
+
+    public String generateServiceAuthToken(){
```
geek-soft: Why user service generates service auth tokens? It does not sounds like users service responsibility. Please find better place for it.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/UserService.java
dharmendrak: renamed

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
dharmendrak: ignore it, removed now, was just debugging

build.gradle
dharmendrak: Rafal k suggested to start release 0.0.1 , now we have 0.0.2 after a fix by Tim. Rafal also want to add some more api in this for their use like 'details'

src/main/java/uk/gov/hmcts/cmc/claimstore/services/UserService.java
dharmendrak: it should be cached one, first time it caches call original bean and later it uses from it's cache. tested locally while debug using docker

src/main/java/uk/gov/hmcts/cmc/claimstore/services/UserService.java
dharmendrak: moved to new service

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ServiceAuthProviderService.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.hmcts.cmc.claimstore.services;
+
+import uk.gov.hmcts.reform.authorisation.generators.AuthTokenGenerator;
+
+public class ServiceAuthProviderService {
+    private final AuthTokenGenerator cachedServiceAuthTokenGenerator;
+
+    public ServiceAuthProviderService(final AuthTokenGenerator cachedServiceAuthTokenGenerator) {
```
geek-soft: Any reason you created service? That does not give any value and makes traversing code more difficult. We need to stop creating useless classes.

src/test/java/uk/gov/hmcts/cmc/claimstore/services/ServiceAuthProviderServiceTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+import uk.gov.hmcts.reform.authorisation.generators.AuthTokenGenerator;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ServiceAuthProviderServiceTest {
```
geek-soft: For me personally that test is useless the same way as `ServiceAuthProviderService` class.

src/main/java/uk/gov/hmcts/cmc/claimstore/config/ServiceTokenGeneratorConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+public class ServiceTokenGeneratorConfiguration {
+
+    @Bean
+    @ConditionalOnMissingBean(name = "authTokenGenerator")
+    public AuthTokenGenerator authTokenGenerator(
+            @Value("${idam.s2s-auth.totp_secret}") final String secret,
+            @Value("${idam.s2s-auth.microservice}") final String microService,
+            final ServiceAuthorisationApi serviceAuthorisationApi
+    ) {
+        return new ServiceAuthTokenGenerator(secret, microService, serviceAuthorisationApi);
```
geek-soft: Why not to have single bean factory method that always returns cached instance? I guess we never going to use the other one (non cached).

dharmendrak: in a way you are right, can have anonymous class for it with in cached bean

build.gradle
```diff
@@ -123,6 +123,7 @@ dependencies {
 
   compile group: 'io.github.openfeign', name: 'feign-httpclient', version: '9.5.0'
   compile group: 'uk.gov.hmcts.reform', name: 'document-management-client', version: '1.0.2'
+  compile group: 'uk.gov.hmcts.reform', name: 'service-auth-provider-client', version: '0.0.4'
```
timja-kainos: this is missing a heathcheck, preferably the library would provide a healthindicator and all consumers would automatically get it

build.gradle
```diff
@@ -123,6 +123,7 @@ dependencies {
 
   compile group: 'io.github.openfeign', name: 'feign-httpclient', version: '9.5.0'
```
timja-kainos: your story had no information or acceptance criteria on it, so posting it here doesn't give much value, can you add a description and a/c please

build.gradle
dharmendrak: is it something new requirement, none of client I feel provide health indicator

timja-kainos: pdf-service-client provides a health method and in the claim-store we have a health indicator, that's the minimum needed.
but it would be much better if the client provides the health indicator.

This was missed on the new clients but it is a requirement that we have a healthcheck for each service we talk to.

geek-soft: Good call @timja-kainos 👍

geek-soft:
Closing as it will be introduced to master as https://github.com/hmcts/cmc-claim-store/pull/117

PR #120, opened 2017-12-07
ROC-2703: Introduce FullDefenceResponse

radoslawn-kainos:

[ROC-2703](https://tools.hmcts.net/jira/browse/ROC-2703)

### Change description ###

Introduce FullDefenceResponse as (the only, as of now) subtype of
Response, and use it in place of Response.

```
[x] Yes
[ ] No
```

### DIFF
--- a/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/FullDefenceResponse.java
+++ b/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/FullDefenceResponse.java
@@ -1,0 +1,67 @@
+package uk.gov.hmcts.cmc.domain.models;
+
+import org.hibernate.validator.constraints.NotBlank;
+import uk.gov.hmcts.cmc.domain.models.legalrep.StatementOfTruth;
+import uk.gov.hmcts.cmc.domain.models.party.Party;
+
+import java.util.Objects;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
+public class FullDefenceResponse extends Response {
+
+    public enum DefenceType {
+        DISPUTE,
+        ALREADY_PAID
+    }
+
+    @NotNull
+    private final DefenceType defenceType;
+
+    @NotBlank
+    @Size(max = 99000)
+    private final String defence;
+
+    public FullDefenceResponse(
+        FreeMediationOption freeMediation,
+        MoreTimeNeededOption moreTimeNeeded,
+        Party defendant,
+        StatementOfTruth statementOfTruth,
+        DefenceType defenceType,
+        String defence
+    ) {
+        super(freeMediation, moreTimeNeeded, defendant, statementOfTruth);
+        this.defenceType = defenceType;
+        this.defence = defence;
+    }
+
+    public DefenceType getDefenceType() {
+        return defenceType;
+    }
+
+    public String getDefence() {
+        return defence;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+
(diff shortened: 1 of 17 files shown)


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/FullDefenceResponse.java
```diff
@@ -0,0 +1,67 @@
+package uk.gov.hmcts.cmc.domain.models;
```
timja-kainos: doesn't there need to be a database migration for this to work?

radoslawn-kainos: Yeah, I realised this just after creating this PR. Will add DB migration and update it.

timja-kainos: If possible the DB migration should come in a separate PR with the frontend being already updated to send the data.

Our new environment that we are working towards requires this and we should start getting into the practice of it.

Alternatively you can not do this and we can do it the harder way when we need to, up to you?

geek-soft: @timja-kainos  I have added DB migration, please have a look with your experienced eyes.

src/main/resources/db/migration/V2017_12_08_1132__Prepare_response_structure_for_subclassing.sql
```diff
@@ -0,0 +1,36 @@
...
+  claim
+SET
+  response = JSONB_SET(response, '{defenceType}',  '"ALREADY_PAID"' :: JSONB)
+WHERE response IS NOT NULL AND jsonb_extract_path_text(response, 'response') = 'OWE_ALL_PAID_ALL';
+
+-- Remove old response property
+WITH subquery AS (
+    SELECT
+      id,
+      response::JSONB #- '{response}' as migrated_response
```
timja-kainos: why are you doing this?

geek-soft: This is to remove old property with `OWE_NONE` and `OWE_ALL_PAID_ALL` values. These has been migrated to `defenceType`.

geek-soft: We had:

```
response: "OWE_NONE"
```

and now we have:

```
responseType: "FULL_DEFENCE"
defenceType: "DISPUTE"
```

geek-soft: We had:

```
response: "OWE_ALL_PAID_ALL"
```

and now we have:

```
responseType: "FULL_DEFENCE"
defenceType: "ALREADY_PAID"
```

timja-kainos: why!!!!
https://github.com/hmcts/cmc-claim-store/blob/master/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Response.java#L48
stupid sneaky JsonProperty rename!

timja-kainos: that confused me lots

geek-soft: Agree, that JSON prop rename is rubbish, confused me many time as well.

timja-kainos: I thought the response was the response but that goes in the defence field...

PR #126, opened 2017-12-15
ROC-2572: add GET /interest/calculate endpoint

yarpo:
https://tools.hmcts.net/jira/browse/ROC-2572

### Change description ###
Added edpoint to calculate interests

```
[ ] Yes
[x] No
```


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/InterestAmount.java
```diff
@@ -0,0 +1,24 @@
+package uk.gov.hmcts.cmc.domain.models;
+
+public class InterestAmount {
+
+    public static final int FACTOR = 100;
+
+    private double amount;
```
geek-soft: We should use `BigDecimal` type to reflect the value.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/InterestAmount.java
```diff
@@ -0,0 +1,24 @@
+package uk.gov.hmcts.cmc.domain.models;
+
+public class InterestAmount {
```
geek-soft: Please make the class immutable, no setter please.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CalculateInterestAmountTest.java
```diff
@@ -0,0 +1,42 @@
...
+        MvcResult result = makeRequest("2010-10-10", "2010-10-16", 30, 8000)
+            .andExpect(status().isOk())
+            .andReturn();
+
+        assertThat(deserializeObjectFrom(result, InterestAmount.class))
+            .extracting(InterestAmount::getAmount)
+            .isEqualTo(39.45);
+    }
+
+    protected ResultActions makeRequest(String fromDate, String toDate, double rate, double amount) throws Exception {
```
geek-soft: I guess that could be private.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/InterestRatesController.java
```diff
@@ -0,0 +1,41 @@
...
+        this.interestRateService = interestRateService;
+    }
+
+    @GetMapping("/calculate")
+    @ApiOperation("Calculates interest rate")
+    public InterestAmount calculateInterest(
+        @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) @RequestParam("from_date") LocalDate from,
+        @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) @RequestParam("to_date") LocalDate to,
+        @RequestParam("rate") double rate,
+        @RequestParam("amount") double amount
```
geek-soft: We should use BigDecimal here as well.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CalculateInterestAmountTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.test.web.servlet.MvcResult;
+import org.springframework.test.web.servlet.ResultActions;
+import uk.gov.hmcts.cmc.claimstore.BaseSaveTest;
+import uk.gov.hmcts.cmc.domain.models.InterestAmount;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+public class CalculateInterestAmountTest extends BaseSaveTest {
```
geek-soft: We should have a test documenting the behaviour when not all parameters required for calculation are provided. We expect some sort of 4xx error at that point.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/InterestRateService.java
```diff
@@ -0,0 +1,42 @@
...
+import java.time.temporal.ChronoUnit;
+
+@Component
+public class InterestRateService {
+
+    public InterestAmount calculateRate(LocalDate fromDate, LocalDate toDate, double rate, double amount) {
+        validateNumberIsNonNegative(rate);
+        validateNumberIsNonNegative(amount);
+
+        final double noOfDays = calculateNumberOfDays(fromDate, toDate);
```
geek-soft: What if number of days is negative e.g somebody passed end date is before start date? We should probably reject such calculation as that indicates programmatic error.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/InterestRateService.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.cmc.domain.models.InterestAmount;
+
+import java.time.LocalDate;
+import java.time.temporal.ChronoUnit;
+
+@Component
+public class InterestRateService {
+
+    public InterestAmount calculateRate(LocalDate fromDate, LocalDate toDate, double rate, double amount) {
```
geek-soft: I would expect this service to return BigDecimal and then controller wraps value with `InterestAmount` which is RESTful data object. The `InterestRateService` should be used in all other places where we need to recalculate interest.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/InterestRateService.java
```diff
@@ -0,0 +1,42 @@
+package uk.gov.hmcts.cmc.claimstore.services;
+
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.cmc.domain.models.InterestAmount;
+
+import java.time.LocalDate;
+import java.time.temporal.ChronoUnit;
+
+@Component
+public class InterestRateService {
```
geek-soft: I guess this service should be used in all places where interest calculation is needed so we have one component responsible for calculation. Have you tried to find other places where we calculate interest and reuse this service in such places?

src/main/java/uk/gov/hmcts/cmc/claimstore/services/InterestRateService.java
```diff
@@ -0,0 +1,42 @@
...
+        if (fromDate == null || toDate == null) {
+            throw new IllegalArgumentException("fromDateDate or toDateDate is null");
+        }
+
+        if (fromDate.isAfter(toDate)) {
+            throw new IllegalStateException("fromDateDate must not be after toDateDate");
+        }
+    }
+
+    private void validateNumberIsNonNegative(final double value) {
```
geek-soft: How that translates to response code? Such exceptions should be mapped to one of the 4xx responses, please double check we are not returning 500 for the error, that would not be correct.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/InterestRateService.java
yarpo: I wanted to keep this PR as small as possible. Only new endpoint and tests for that.

Then next one - using it everywhere, where it should be used.

And one more for FE to call API whenever it should be calculated.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/InterestRateService.java
yarpo: Please see `validatePeriod(fromDate, toDate);`

```
private double calculateNumberOfDays(final LocalDate fromDate, final LocalDate toDate) {
    validatePeriod(fromDate, toDate);
    return fromDate.until(toDate, ChronoUnit.DAYS);
}
```

rlewan:
I'm ok with using the new service everywhere being done as a separate PR. However, all instances of `double` should be replaced with `BigDecimal`.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CalculateInterestAmountTest.java
yarpo: added

rlewan:
@yarpo @geek-soft I think the correct way to do this is to validate the inputs correctly on the endpoint level. Then we don't need to throw a HTTP specific exception from the service.

I talked to Patryk about it and he told me there is an issue with Spring not enforcing the presence of input parameters.. Would be nice if that could be solved. I'm guessing the problem is that with `RequestParam(required = true` (which is the default), Spring will check only for the existence of query param, e.g. `?from_date=`, but will not check if it has any value assigned.

domain-model/src/test/java/uk/gov/hmcts/cmc/domain/models/InterestAmountTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class InterestAmountTest {
+
+    @Test
+    public void valueOfReturnObjectWithAmountEq0() {
```
geek-soft: I guess we do not need this. It sort of suggests that there is special handling of zero values while only job that class is doing is you rounding half UP which might be documented in one tests case.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/InterestRateService.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    private long calculateNumberOfDays(LocalDate fromDate, LocalDate toDate) {
+        validatePeriod(fromDate, toDate);
+        return fromDate.until(toDate, ChronoUnit.DAYS);
+    }
+
+    private void validatePeriod(LocalDate fromDate, LocalDate toDate) {
+        if (fromDate == null || toDate == null) {
+            throw new BadRequestException("fromDateDate or toDateDate is null");
```
geek-soft: The message is misspelled here and in if statement below.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/InterestRateService.java
```diff
@@ -0,0 +1,42 @@
...
+@Component
+public class InterestRateService {
+
+    public BigDecimal calculateRate(LocalDate fromDate, LocalDate toDate, BigDecimal rate, BigDecimal amount) {
+        validateVarIsNonNegativeBigDecimal(rate);
+        validateVarIsNonNegativeBigDecimal(amount);
+
+        long noOfDays = calculateNumberOfDays(fromDate, toDate);
+
+        return BigDecimal.valueOf((amount.doubleValue() * noOfDays * rate.doubleValue()) / (365 * 100));
```
geek-soft: We should use `BigDecimal` methods for calculation. The whole point of this story is to increase correctness of the calculation result, making calculation on flouting point numbers does not guarantee that.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/InterestRateService.java
geek-soft: I just looked into the code and found `uk.gov.hmcts.cmc.claimstore.services.interest.InterestCalculationService#calculateInterest ` method which seems to do exactly what this service is going. Please reuse existing one or merge these two.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/InterestRateService.java
```diff
@@ -0,0 +1,46 @@
...
+    public BigDecimal calculateRate(LocalDate fromDate, LocalDate toDate, BigDecimal rate, BigDecimal amount) {
+        validateVarIsNonNegativeBigDecimal(rate);
+        validateVarIsNonNegativeBigDecimal(amount);
+
+        long noOfDays = calculateNumberOfDays(fromDate, toDate);
+
+        return amount
+            .multiply(rate)
+            .multiply(BigDecimal.valueOf(noOfDays))
+            .divide(BigDecimal.valueOf(365 * 100), 2, RoundingMode.HALF_UP);
```
jenkins-reform-hmcts: MINOR Cast one of the operands of this multiplication operation to a "long". [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AS2184)

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/InterestAmount.java
```diff
@@ -0,0 +1,17 @@
...
+
+    private BigDecimal amount;
+
+    public InterestAmount(BigDecimal amount) {
+        this.amount = amount;
+    }
+
+    public BigDecimal getAmount() {
+        return amount.setScale(2, RoundingMode.HALF_UP);
```
rlewan: Since the class is immutable now we can move this call to constructor.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CalculateInterestAmountTest.java
yarpo: but they are 5xx errors now as we do not map these exceptions to different status codes.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/InterestAmount.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.hmcts.cmc.domain.models;
+
+import java.math.BigDecimal;
+
+public class InterestAmount {
+
+    private BigDecimal amount;
```
geek-soft: Please make field final.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CalculateInterestAmountTest.java
```diff
@@ -0,0 +1,72 @@
...
+            .andExpect(status().isOk())
+            .andReturn();
+
+        InterestAmount obj = deserializeObjectFrom(result, InterestAmount.class);
+        assertThat(obj.getAmount()).isEqualTo(BigDecimal.valueOf(39.45));
+    }
+
+    @Test
+    public void shouldReturnBadRequestWhenDateFromIsEmpty() throws Exception {
+        makeRequest("", "2010-10-10", 30, 8000).andExpect(status()
```
geek-soft: I would love to break like before `.andExpect` here and in other places.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CalculateInterestAmountTest.java
```diff
@@ -0,0 +1,72 @@
...
+        MvcResult result = makeRequest("2010-10-10", "2010-10-16", 30, 8000)
+            .andExpect(status().isOk())
+            .andReturn();
+
+        InterestAmount obj = deserializeObjectFrom(result, InterestAmount.class);
+        assertThat(obj.getAmount()).isEqualTo(BigDecimal.valueOf(39.45));
+    }
+
+    @Test
+    public void shouldReturnBadRequestWhenDateFromIsEmpty() throws Exception {
```
geek-soft: The test description does not match the behaviour. Either change it to say that internal server error is returned or add validation in the controller to validate input one more time and throw BadRequestException.

The second approach is better but I am able to accept 500 for this internal API.

src/test/java/uk/gov/hmcts/cmc/claimstore/services/interest/InterestCalculationServiceInterestUpToNowTest.java
```diff
@@ -53,7 +53,7 @@ public void interestUpToNowFrom12345PoundsOf75PercentFor500DaysShouldEqualExpect
         BigDecimal calculated = service.calculateInterestUpToNow(
             MORE_POUNDS, SEVENTY_FIVE_PERCENT_INTEREST, FIVE_HUNDRED_DAYS_AGO
         );
-        assertThat(calculated).isEqualByComparingTo("12674.81");
+        assertThat(calculated).isEqualByComparingTo("12683.22");
```
geek-soft: It is Christmas time, you adding more money to people :)

yarpo: ho ho ho ;)

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CalculateInterestAmountTest.java
yarpo: I'll to the 2nd one

yarpo: done

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CalculateInterestAmountTest.java
yarpo: done

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/InterestRatesController.java
```diff
@@ -0,0 +1,47 @@
...
+    @ApiOperation("Calculates interest rate")
+    public InterestAmount calculateInterest(
+        @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) @RequestParam("from_date") LocalDate from,
+        @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) @RequestParam("to_date") LocalDate to,
+        @RequestParam("rate") BigDecimal rate,
+        @RequestParam("amount") BigDecimal amount
+    ) {
+        try {
+            return new InterestAmount(interestCalculationService.calculateInterest(amount, rate, from, to));
+        } catch (IllegalArgumentException e) {
```
jenkins-reform-hmcts: MAJOR Either log or rethrow this exception. [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AS1166)

src/test/java/uk/gov/hmcts/cmc/claimstore/services/interest/InterestCalculationServiceTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class InterestCalculationServiceTest {
+
+    private static final LocalDate START_DATE = LocalDate.now();
+    private static final BigDecimal NEGATIVE = BigDecimal.valueOf(-12);
+
+    private InterestCalculationService service = new InterestCalculationService(Clock.systemDefaultZone());
+
+    @Test(expected = IllegalArgumentException.class)
+    public void shouldIllegalArgumentExceptionWhenAmountIsNegative() {
```
geek-soft: Please correct to be `shouldThrow...`. A `shouldIllegal...` does not read well.

yarpo: changed

PR #137, opened 2018-01-08
ROC-2572: calculate interest on server and returned in response

yarpo:
https://tools.hmcts.net/jira/browse/ROC-2572

### Change description ###
Add 2 fields to Claim class and populate them before returning object from API

```
[ ] Yes
[x] No
```


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
```diff
@@ -162,6 +169,14 @@ public LocalDateTime getSettlementReachedAt() {
         return Optional.ofNullable(sealedClaimDocumentSelfPath);
     }
 
+    public BigDecimal getTotalAmountTillToday() {
```
timja-kainos: Looking at the code that had to be added to support this I think it would be cleaner if these get methods did the calculation and not the service methods,

i.e. things like this aren't nice at all:
https://github.com/hmcts/cmc-claim-store/pull/137/files#diff-55cdaa2fa8cc537e615c8ca0118170a5R98

and the copy method isn't ideal.

Thoughts?

yarpo: Yeah, I agree. I am more than open for suggestions.

I'd:
1. move calculations to some separate service.
2. leave `copy` method as it.

More unit tests and un-ignore one int test (that was failing).

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/SaveDefendantResponseTest.java
```diff
@@ -60,6 +61,7 @@ public void shouldReturnNewlyCreatedDefendantResponse() throws Exception {
             .contains(Optional.of(response));
     }
 
+    @Ignore
```
rlewan: Are you planning to enable that test back later?

yarpo: Yes

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -41,51 +48,75 @@ public ClaimService(
...
+        }
+
+        return optionalClaim;
     }
 
     public List<Claim> getClaimByExternalReference(String externalReference, String authorisation) {
         String submitterId = userService.getUserDetails(authorisation).getId();
-        return claimRepository.getByExternalReference(externalReference, submitterId);
+        return claimRepository.getByExternalReference(externalReference, submitterId)
+            .stream().map((claim) -> calculateAndPopulateTotalAmount(claim)).collect(Collectors.toList());
```
jenkins-reform-hmcts: MINOR Replace this lambda with a method reference. [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AS1612)
MINOR Remove the parentheses around the "claim" parameter [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AS1611)

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -41,51 +48,75 @@ public ClaimService(
...
+            claimRepository
+                .getById(claimId)
+                .orElseThrow(() -> new NotFoundException("Claim not found by id " + claimId))
+        );
     }
 
     public List<Claim> getClaimBySubmitterId(String submitterId) {
-        return claimRepository.getBySubmitterId(submitterId);
+        return claimRepository.getBySubmitterId(submitterId).stream()
+            .map((claim) -> calculateAndPopulateTotalAmount(claim)).collect(Collectors.toList());
```
jenkins-reform-hmcts: MINOR Replace this lambda with a method reference. [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AS1612)
MINOR Remove the parentheses around the "claim" parameter [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AS1611)

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
geek-soft: Agree that these getters could calculate values the same way as we used to do it in the frontend.

That would mean that `uk.gov.hmcts.cmc.claimstore.services.interest.InterestCalculationService#calculateInterest` becomes a dependency of `uk.gov.hmcts.cmc.domain.models.Claim` however we can accept it.

That would remove need for custom `copy` method.

rlewan: @timja-kainos I had a chat about this with Patryk yesterday and calculating in getter methods was his first thought, but that would make a dependency from the models module to the core module.

I'm not sure now how we intended to use these modules. If the models module is for keeping entities only then this dependency wouldn't be very nice.

But if we were to do it as DDD suggests, i.e. to keep the domain model layer separate, then it shouldn't be an issue. We would need to extract all domain related logic into the models module (or a separate domain focused one) though. Maybe not all in the first step, but that could be one way for us to proceed.

Any more thoughts @geek-soft ?

geek-soft: I would be keen to keep calculator class (e.g. service) close to the model in case we need to extract domain classes to separate logical module.

I understand that keeping everything in the domain might prove to be difficult in the future and it is not necessarily what we would like to do especially when it requires some HTTP requests to external services e.g. to check UK bank holidays.

I think we will be iterating / improving this over the time, for now keeping total value calculation in the claim class makes sense.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/amount/TotalAmountCalculator.java
```diff
@@ -0,0 +1,103 @@
...
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.time.Duration;
+import java.time.LocalDate;
+import java.util.Optional;
+
+import static java.math.BigDecimal.valueOf;
+import static java.util.Objects.requireNonNull;
+
+public class TotalAmountCalculator {
```
rlewan: From what I can see this class calculates other things in addition to total amounts. What do you think about changing the name to be less specific?

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
```diff
@@ -13,6 +16,7 @@
 import static uk.gov.hmcts.cmc.domain.utils.ToStringStyle.ourStyle;
 
 @JsonInclude(JsonInclude.Include.NON_ABSENT)
+@JsonIgnoreProperties(value = {"totalAmountTillToday", "totalAmountTillDateOfIssue"}, allowGetters = true)
```
rlewan: Oh, did you manage to get this working in the end?

domain-model/src/test/java/uk/gov/hmcts/cmc/domain/amount/TotalAmountCalculatorTest.java
```diff
@@ -0,0 +1,148 @@
...
+        ).isEqualTo(format(new BigDecimal(219.18)));
+    }
+
+    @Test
+    public void totalTillTodayShouldReturnEmptyOptionalWhenClaimHasAmountDifferentThanAmountBreakdown() {
+        assertThat(TotalAmountCalculator.totalTillToday(claimWithAmountRange())).isEqualTo(Optional.empty());
+    }
+
+    @Test
+    public void totalTillTodayShouldReturnClaimAmountPlusFeeWhenNoInterestWhenNoIntrest() {
```
rlewan: `WhenNoIntrest` has a typo and is a duplicate, can you please remove it?

domain-model/src/test/java/uk/gov/hmcts/cmc/domain/amount/TotalAmountCalculatorTest.java
```diff
@@ -0,0 +1,148 @@
...
+                    .withFeeAmount(TWENTY_POUNDS_IN_PENNIES)
+                    .build()
+            ).build();
+
+        assertThat(TotalAmountCalculator.totalTillToday(claimNoInterest))
+            .isEqualTo(Optional.of(format(new BigDecimal(60))));
+    }
+
+    @Test
+    public void totalTillDateOfIssueShouldReturnClaimAmountPlusWhenNoInterestWhenNoIntrest() {
```
rlewan: Same here.

rlewan: Also - "should return claim amount plus"...?

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/amount/TotalAmountCalculator.java
yarpo: `AmountCalculator`?

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
yarpo: finally yes. This let us return claim with these fields and deserialise JSON into Java object without taking these fields into account.

domain-model/src/test/java/uk/gov/hmcts/cmc/domain/amount/TotalAmountCalculatorTest.java
yarpo: renamed. 

Now:
`totalTillTodayShouldReturnClaimAmountPlusFeeWhenNoInterestWhenNoInterest`
and 
`totalTillDateOfIssueShouldReturnClaimAmountPlusFeeWhenNoInterestWhenNoInterest`

domain-model/src/test/java/uk/gov/hmcts/cmc/domain/amount/TotalAmountCalculatorTest.java
yarpo: done

domain-model/src/test/java/uk/gov/hmcts/cmc/domain/amount/TotalAmountCalculatorTest.java
rlewan: I can still see duplicated `WhenNoInterest` in those examples 😛

yarpo: renamed (WhenNoInterestWhenNoInterest -> WhenNoInterest)

PR #139, opened 2018-01-10
ROC-3003: Call CCD save only if the transaction is committed

sabahirfan:
[ROC-3003](https://tools.hmcts.net/jira/browse/ROC-3003)

### Change description ###
This PR fixes the scenario where save claim transaction rolls back but the CCD uploader event listener does not rolls back. The fix will invoke CCD uploader after commit of the transaction.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/events/ccd/CoreCaseDataUploader.java
```diff
@@ -21,7 +21,7 @@ public CoreCaseDataUploader(CoreCaseDataService coreCaseDataService) {
         this.coreCaseDataService = coreCaseDataService;
     }
 
-    @EventListener
+    @TransactionalEventListener
```
rlewan: That's cool, didn't know about it 👍

PR #152, opened 2018-01-17
Fix nullpointer when no interest

timja-kainos:


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/InterestDateMapper.java
```diff
@@ -10,6 +10,10 @@
 
     @Override
     public CCDInterestDate to(InterestDate interestDate) {
+        if (interestDate == null || interestDate.getType() == null) {
+            return null;
```
rlewan: Can we return `Optional` here?

timja-kainos: Not sure if we could get any benefit here, the mappers aren't invoked directly and the pattern has been to return null from there if they could be empty.

rlewan: I had a look into the `Mappper` interface and you're right.

Another matter - `|| interestDate.getType() == null` - it seems to me an interest date without a type is an application error, valid object should be sent to us from frontend. I would prefer not to hid the problem here.

timja-kainos: not a frontend problem from what I can tell so far, but no idea why its set. Will dig a bit more but would like to merge this if we can't find it soon

PR #160, opened 2018-01-24
ROC-3126: Revise claim PDF templates

radoslawn-kainos:

[ROC-3126](https://tools.hmcts.net/jira/browse/ROC-3126)

### Change description ###

Revise claim PDF templates

```
[ ] Yes
[x] No
```


src/main/resources/staff/templates/document/sealedClaim.html
```diff
@@ -88,25 +88,24 @@
...
         <p>{{ claimant.fullName }}’s claim</p>
       </td>
     </tr>
     <tr>
-      <td colspan="2"><p><strong>Submitted on: </strong> {{ claim.submittedOn }}</p></td>
+      <td><p><strong>Submitted on: </strong> {{ claim.submittedOn }}</p></td>
     </tr>
     <tr>
-      <td colspan="2"><p><strong>Issued on: </strong> {{ claim.issuedOn }}</p></td>
+      <td><p><strong>Issued on: </strong> {{ claim.issuedOn }}</p></td>
```
timja-kainos: beware long months, this was changed to 2 months for September I think. Could you test with September and make sure there's no wrapping please?

radoslawn-kainos: @timja-kainos Why would there be any wrapping introduced here? Removing `colspan` from these cells resulted in there being one less column overall, so the cells that had `colspan="2"` should be as wide now as they were then. Well, at least unless there are some column sizing styles messing it up, but it doesn't seem like it.

timja-kainos: sounds good, was just a little worried, because we had to make changes to make September fit, we first built them in July and may not have tested longer months

PR #164, opened 2018-01-26
ROC-3097 Use external ID for defendant linking

timja-kainos:
and add to case repository

https://tools.hmcts.net/jira/browse/ROC-3024

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/BaseIntegrationTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/BaseIntegrationTest.java
@@ -27,6 +27,7 @@ public abstract class BaseIntegrationTest extends MockSpringTest {
 
     protected static final String SUBMITTER_ID = "123";
     protected static final String DEFENDANT_ID = "555";
+    public static final String BEARER_TOKEN = "Bearer let me in";
 
     @Autowired
     protected MockMvc webClient;
--- a/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/BaseIntegrationTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/BaseIntegrationTest.java
@@ -27,7 +27,7 @@ public abstract class BaseIntegrationTest extends MockSpringTest {
 
     protected static final String SUBMITTER_ID = "123";
     protected static final String DEFENDANT_ID = "555";
-    public static final String BEARER_TOKEN = "Bearer let me in";
+    protected static final String BEARER_TOKEN = "Bearer let me in";
 
     @Autowired
     protected MockMvc webClient;
--- a/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/AcceptOrRejectOfferTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/AcceptOrRejectOfferTest.java
@@ -57,7 +57,7 @@ public class AcceptOrRejectOfferTest extends BaseIntegrationTest {
         );
 
         claim = claimStore.saveClaim(SampleClaimData.builder().build(), SUBMITTER_ID, LocalDate.now());
-        caseRepository.linkDefendant(claim.getExternalId(), "2", BEARER_TOKEN);
+        caseRepository.linkDefendant(claim.getExternalId(), DEFENDANT_ID, BEARER_TOKEN);
         claimStore.saveResponse(claim.getId(), SampleResponse.validDefaults(), DEFENDANT_ID,
             SampleClaim.DEFENDANT_EMAIL);
         prepareDefendantOffer();
--- a/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/AcceptOrRejectOfferTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/AcceptOrRejectOfferTest.java
@@ -57,7 +57,7 @@ public class AcceptOrRejectOfferTest extends BaseIntegrationTest {
         );
 
         claim = claimStore.saveClaim(SampleClaimData.builder().build(), SUBMITTER_ID, LocalDate.now());
-        claimRepository.linkDefendant(claim.getId(), DEFENDANT_ID);
+        caseRepository.linkDefendant(claim.getExternalId(), "2", BEARER_TOKEN);
         claimStore.saveResponse(claim.getId(), SampleResponse.validDefaults(), DEFENDANT_ID,
             SampleClaim.DEFENDANT_EMAIL);
         prepareDefendantOffer();
--- a/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/LinkDefendantToClaimTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/LinkDefendantToClaimTest.java
@@ -18,7 +18,7 @@ public class LinkDefendantToClaimTest extends BaseIntegrationTest {
     public void shouldReturn200HttpStatusAndUpdatedClaimWhenLinkIsSuccessfullySet() throws Exception {
         Claim claim = claimStore.saveClaim(SampleClaimData.builder().build());
 
-        MvcResult result = linkdefendantRequest(claim.getExternalId(), "1")
+        MvcResult result = linkDefendantRequest(claim.getExternalId(), "1")
             .andExpect(status().isOk())
             .andReturn();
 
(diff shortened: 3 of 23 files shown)


src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/AcceptOrRejectOfferTest.java
```diff
@@ -57,7 +57,7 @@ public void beforeEachTest() throws Exception {
         );
 
         claim = claimStore.saveClaim(SampleClaimData.builder().build(), SUBMITTER_ID, LocalDate.now());
-        claimRepository.linkDefendant(claim.getId(), DEFENDANT_ID);
+        caseRepository.linkDefendant(claim.getExternalId(), "2", BEARER_TOKEN);
```
geek-soft: Reusing `DEFENDANT_ID` is more readable. As a reader I am surprised that some random defendant identified by ID equal 2 is used.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/LinkDefendantToClaimTest.java
```diff
@@ -28,14 +29,15 @@ public void shouldReturn200HttpStatusAndUpdatedClaimWhenLinkIsSuccessfullySet()
...
-        long nonExistingClaimId = 900;
+        String nonExistingExternalId = "7d293143-b787-454f-aa8e-2fd69a209e52";
 
-        linkdefendantRequest(nonExistingClaimId, "1")
+        linkdefendantRequest(nonExistingExternalId, "1")
             .andExpect(status().isNotFound());
     }
 
-    private ResultActions linkdefendantRequest(Object claimId, String defendantId) throws Exception {
+    private ResultActions linkdefendantRequest(Object externalId, String defendantId) throws Exception {
```
geek-soft: Please make method name properly capitalised to fix something we missed earlier.

geek-soft: Please also make first arg of type String.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/BaseIntegrationTest.java
```diff
@@ -27,6 +27,7 @@
 
     protected static final String SUBMITTER_ID = "123";
     protected static final String DEFENDANT_ID = "555";
+    public static final String BEARER_TOKEN = "Bearer let me in";
```
geek-soft: Please make it protected to discourage anyone from reusing it outside tests that inherit the base class.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/ResendStaffNotificationsTest.java
```diff
@@ -70,7 +72,7 @@ public void shouldRespond409AndNotProceedForClaimIssuedEventWhenClaimIsLinkedToD
         String event = "claim-issued";
 
         Claim claim = claimStore.saveClaim(SampleClaimData.builder().build());
-        claimRepository.linkDefendant(claim.getId(), "2");
+        caseRepository.linkDefendant(claim.getExternalId(), "2", BEARER_TOKEN);
```
geek-soft: Please use `DEFENDANT_ID` constant here to make code more readable.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/ClaimController.java
```diff
@@ -110,7 +112,7 @@ public Claim requestMoreTimeToRespond(@PathVariable("claimId") Long claimId,
     @GetMapping("/{caseReference}/defendant-link-status")
     @ApiOperation("Check whether a claim is linked to a defendant")
     public DefendantLinkStatus isDefendantLinked(@PathVariable("caseReference") String caseReference) {
-        Boolean linked = claimService.getClaimByReference(caseReference)
+        Boolean linked = claimService.getClaimByReferenceAnonymous(caseReference)
```
geek-soft: I would like to delete this endpoint in the future as it provides minimal value. It does not impact this PR of course.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/LegacyCaseRepository.java
```diff
@@ -12,7 +13,7 @@
 
 @RegisterMapper(ClaimMapper.class)
 @SuppressWarnings("squid:S1214") // Pointless to create class for string statement
-public interface ClaimSearchRepository {
+public interface LegacyCaseRepository {
```
geek-soft: Why do we need that interface which is a copy of `uk.gov.hmcts.cmc.claimstore.repositories.ClaimRepository`? We should not need to change `uk.gov.hmcts.cmc.claimstore.repositories.ClaimRepository` when moving to CCD beside small changes triggered by API change.

Moving methods from one place to another makes no much sense.

I would scrap `LegacyCaseRepository ` and keep using old DBI repository.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/search/CCDCaseRepository.java
```diff
@@ -77,4 +77,14 @@ private void logClaimDetails(List<Claim> dbClaims, List<Claim> ccdClaims) {
 
         return claim;
     }
+
+    @Override
+    public Optional<Claim> linkDefendant(String externalId, String defendantId, String authorisation) {
+        Optional<Claim> claim = legacyCaseRepository.getClaimByExternalId(externalId);
```
geek-soft: That does not look right that CCD implementation uses DB for linking. CCD repository should be used instead here.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/AcceptOrRejectOfferTest.java
timja-kainos: accidental 👍

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/LinkDefendantToClaimTest.java
timja-kainos: hah, may as well

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/ClaimController.java
timja-kainos: needs idam to be fixed for that to happen, now that there's developers on it again we can potentially do that

src/main/java/uk/gov/hmcts/cmc/claimstore/services/search/CCDCaseRepository.java
timja-kainos: this is step 1

timja-kainos: will be a follow up PR to do that

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/LegacyCaseRepository.java
timja-kainos: not sure other than it makes it clearer whats been moved to CCD, 
I tried to move more but I got a cyclic dependency.

Moving the methods allowed me to see a couple of places where we still weren't using the auth.

I don't mind either way

geek-soft: My reasoning is that the code should not be used to track what is left, we have stories for that.

My approach would be extract interface from old repository and start implementing the new one, non implemented methods would throw `NotImplementedException` exception.

Anyway nothing to do with this PR, however I would stop moving things between two repos that play the same role.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/search/CCDCaseRepository.java
geek-soft: In that case please throw `NotImplemenetdException` and disable CCD in dev until fully functional. Providing two implementations in the same way as we have done it for few retrieval endpoints (retrieve both and log) is just unnecessary effort.

timja-kainos: This keeps it functional for now though? 
I'll remove it in a follow up but QA will want CCD enabled for now

PR #165, opened 2018-01-26
ROC-3024 Use external id for all api calls

timja-kainos:


geek-soft:
LGTM, just more integration tests need to be updated to accommodate the API change

PR #166, opened 2018-01-29
Add null checks on CCD retrieval

timja-kainos:
Prod will have some data that could be affected by some of these null
checks, I needed all these null checks to be able to retrieve data in my
local environment

### DIFF
--- a/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/AmountMapper.java
+++ b/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/AmountMapper.java
@@ -45,6 +45,10 @@ public class AmountMapper implements Mapper<CCDAmount, Amount> {
 
     @Override
     public Amount from(CCDAmount ccdAmount) {
+        // After data migration remove this null check and enforce presence
+        if (ccdAmount == null || ccdAmount.getType() == null) {
+            return null;
+        }
 
         switch (ccdAmount.getType()) {
             case RANGE:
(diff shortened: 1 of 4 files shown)


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/ClaimMapper.java
```diff
@@ -93,7 +93,10 @@ private CCDPartyArrayElement mapToValue(CCDParty ccdParty) {
 
     @Override
     public ClaimData from(CCDClaim ccdClaim) {
-        Objects.requireNonNull(ccdClaim, "ccdClaim must not be null");
+        // After data migration remove this null check and enforce presence
+        if (ccdClaim == null || ccdClaim.getClaimants() == null) {
```
sabahirfan: IMO claim or claimants can not be null.

ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/AmountMapper.java
```diff
@@ -45,6 +45,10 @@ public CCDAmount to(Amount amount) {
 
     @Override
     public Amount from(CCDAmount ccdAmount) {
+        // After data migration remove this null check and enforce presence
```
geek-soft: What do we need that for? Amount should always exists.

timja-kainos: this may be one we can get away without, but in my local I have cases without it.
Probably going to defer merging this until we have some sort of agreement on how we do the data migrations

geek-soft: Would be great to investigate why you have it locally and whether we have the same in prod. The only legitimate case I can see right now is the scenario where we changed the field names without reflecting such change in existing data in CCD that causes 'Unrecognised field' error.

timja-kainos:
Closing due to this not being needed urgently for a release, we will fix it when we do a full data migration

PR #167, opened 2018-01-29
ROC-3020 Rename repository classes to avoid confusion

dharmendrak:
This PR is to achieve the following class structure.

nomnoml-2

This PR made following changes,
1) All search methods extracted from claim repository to legacySearchRepository are now merged back in claim repository.
2) Renamed existing CaseRepository implementer classes to DBCaseReposiory , CCDCaseRepository


src/main/java/uk/gov/hmcts/cmc/claimstore/services/search/CCDCaseDBI.java
```diff
@@ -14,28 +14,28 @@
 
 import static java.lang.String.format;
 
-@Service("caseRepository")
+@Service("caseDBI")
 @ConditionalOnProperty(prefix = "core_case_data", name = "api.url")
-public class CCDCaseRepository implements CaseRepository {
-    private final Logger logger = LoggerFactory.getLogger(CCDCaseRepository.class);
+public class CCDCaseDBI implements CaseDBI {
```
timja-kainos: Repository is the spring boot name for the data access class

I would prefer:
keep ClaimRepository and revert this back to what it currently was

dharmendrak: we will be discarding all claim names once we move to CCD, Claim->Case, ClaimService ->CaseService, and so fourth.

timja-kainos: At that point ClaimRepository will be deleted

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/ClaimStore.java
```diff
@@ -25,10 +24,7 @@
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
     @Autowired
-    private ClaimRepository claimRepository;
-
-    @Autowired
-    private LegacyCaseRepository legacyCaseRepository;
+    private CaseDBI caseDBI;
```
timja-kainos: where has DBI come from?
Its not a spring term as far as I'm aware,
we use repository

geek-soft:
Another possible variant: 

nomnoml-3

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -22,7 +22,7 @@
 
 @Service
 @ConditionalOnProperty(prefix = "core_case_data", name = "api.url")
-public class CCDClaimSearchRepository {
+public class CCDCaseApi {
```
geek-soft: The name is a bit unfortunate but I am happy o accept it for now as in the future the implementation is likely to be moved to `CCDCaseRepository`.

PR #170, opened 2018-01-30
RDM-1511 Point towards CCD on CNP

SiIrving:
Looking to point your app to CCD on CNP just for a quick 'sanity check' if thats OK?

### DIFF
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -29,7 +29,7 @@ variable "ccd-url" {
 }
 
 variable "frontend-url" {
-  default = "http://ccd-data-store-api-sandbox.service.core-compute-sandbox.internal"
+  default = "https://www.nonprod.moneyclaim.platform.hmcts.net"
 }
 
 variable "database-name" {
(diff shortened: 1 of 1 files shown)


infrastructure/variables.tf
```diff
@@ -25,11 +25,11 @@ variable "s2s-url" {
 }
 
 variable "ccd-url" {
-  default = "https://case-data-app.test.ccd.reform.hmcts.net:4481"
+  default = "http://ccd-data-store-api-sandbox.service.core-compute-sandbox.internal"
```
timja-kainos: move this to main.tf, look at the existing samples like the pdf service

infrastructure/variables.tf
```diff
@@ -25,11 +25,11 @@ variable "s2s-url" {
...
 }
 
 variable "frontend-url" {
-  default = "https://www.nonprod.moneyclaim.platform.hmcts.net"
+  default = "http://ccd-data-store-api-sandbox.service.core-compute-sandbox.internal"
```
timja-kainos: why are you changing this?

infrastructure/variables.tf
SiIrving: Are you sure? This is where you had it originally? Why do you want me to change this away from where you had it?

PR #173, opened 2018-01-31
ROC-3024 Link defendant

timja-kainos:
https://tools.hmcts.net/jira/browse/ROC-3024


src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -65,22 +74,98 @@ public CCDCaseApi(
...
+        User user = userService.authenticateAnonymousCaseWorker();
+        CaseDetails caseResult = searchByExternalId(authorisation, externalId)
+            .orElseThrow(() -> new CoreCaseDataStoreException("Claim not found by external ID"));
+
+        caseAccessApi.grantAccessToCase(user.getAuthorisation(),
+            authTokenGenerator.generate(),
+            user.getUserDetails().getId(),
+            JURISDICTION_ID,
+            CASE_TYPE_ID,
+            caseResult.getId() + "",
```
rlewan: What do you think about just calling `.toString()` on the ID explicitly?

timja-kainos: don't mind, this is CCD messing up their datatypes :(

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -65,22 +74,98 @@ public CCDCaseApi(
...
+                    userDetails.getId(),
+                    JURISDICTION_ID,
+                    CASE_TYPE_ID,
+                    caseId
+                ).getData()
+            )
+        );
+    }
+
+    private Optional<CaseDetails> searchByExternalId(String authorisation, String externalId) {
```
rlewan: We have two methods `searchByExternalId` and `getByExternalId` where the latter calls the former and additionally maps a CCD object to Claim Store domain object. Would you mind renaming the methods so that intents are expressed more clearly?

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -65,22 +74,103 @@ public CCDCaseApi(
...
+        return getCaseDetailsByExternalId(authorisation, externalId)
+            .map(CaseDetails::getData)
+            .map(this::convertToCCDCase)
+            .map((caseMapper::from));
+    }
 
-        if (claims.size() > 1) {
+    public Optional<Claim> linkDefendant(String externalId, String defendantId, String authorisation) {
+        User user = userService.authenticateAnonymousCaseWorker();
+        Optional<CaseDetails> optionalCaseDetails = getCaseDetailsByExternalId(authorisation, externalId);
```
geek-soft: Will that give any result at all? I guess authorization header comes from the defendant which does not have access to case at this point yet.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/OffersController.java
```diff
@@ -109,9 +109,8 @@ private void assertActionIsPermittedFor(Claim claim, MadeBy party, String author
         if (party.equals(MadeBy.CLAIMANT)) {
             authorisationService.assertIsSubmitterOnClaim(claim, userDetails.getId());
         }
-        if (party.equals(MadeBy.DEFENDANT)) {
-            authorisationService.assertIsDefendantOnClaim(claim, userDetails.getId());
-        }
+        // We have no way to check if a user is a defendant on the claim other than CCD allowing the save
```
rlewan: In this case we should just have a single `isPartyOnClaim` or `hasAccessToClaim` method in the auth service. Please update this.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -65,22 +74,103 @@ public CCDCaseApi(
...
+                user.getUserDetails().getId(),
+                JURISDICTION_ID,
+                CASE_TYPE_ID,
+                caseDetails.getId().toString(),
+                new UserId(defendantId)
+            );
+
+            return Optional.of(caseMapper.from(convertToCCDCase(caseDetails.getData())));
+        }
+        return Optional.empty();
```
rlewan: We just silently return a blank optional if linking fails because given case was not found. I think we should throw an exception in such case.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/OffersController.java
timja-kainos: IMO we can't do any of these checks and we should just delete this service, its up to CCD to do authorisation now. but I think this can be deleted later

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -185,6 +185,8 @@ public void saveCountyCourtJudgment(String externalId, CountyCourtJudgment count
     }
 
     public void saveDefendantResponse(long claimId, String defendantId, String defendantEmail, Response response) {
+        // When this is saved in CCD ensure a Forbidden response is returned to the client if they
```
rlewan: What do you think about adding a story for this/updating an existing story to do this/at least adding a `TODO` tag?

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
timja-kainos: up a level it throws a not found exception

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
timja-kainos: its right on a line that has to change when we save the defendant response

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
timja-kainos: nope, was just fixing that

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/SaveCoreCaseDataService.java
```diff
@@ -55,7 +69,20 @@ public CaseDetails save(
             .data(toJson(ccdCase))
             .build();
 
-        return submit(authorisation, eventRequestData, caseDataContent, represented);
+        CaseDetails caseDetails = submit(authorisation, eventRequestData, caseDataContent, represented);
+
+        if (letterHolderId != null) {
```
rlewan: Will there be cases where null `letterHolderId` will be passed to the `save` method? If yes then I think we should have 2 methods -- e.g. `save` as it is now and `saveWithLetterHolder`, which calls the first one and then performs the letter holder linking logic. I think it's best to avoid methods which have "optional" arguments and require nulls to be passed in. What do you think?

timja-kainos: its not that, its represented or not currently, I didn't want to check the represented flag as they are building a defendant journey, but if I checked the represented flag would you be happier with that?

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/OffersController.java
rlewan: Until we remove the service and let CCD do authorisation, can you please rename the method so that it's not related to claimant only?

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
rlewan: I still believe throwing the exception is the responsibility of this method.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/OffersController.java
timja-kainos: removed method as I can't see a way where it still makes sense

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
timja-kainos: You think a NotFoundException should come from here?
Happy to do but normally would throw a bit higher.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/SaveCoreCaseDataService.java
rlewan: I'd prefer 2 methods but a boolean will make me slightly happier than with as it is now 🙂

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/GetClaimByClaimReferenceFromCoreCaseDataStoreTest.java
```diff
@@ -26,6 +27,7 @@
         "document_management.api_gateway.url=false"
     }
 )
+@Ignore
```
rlewan: Could we please add a comment which says why the test is ignored?

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -65,45 +76,125 @@ public CCDCaseApi(
...
+            authTokenGenerator.generate(),
+            anonymousCaseWorker.getUserDetails().getId(),
+            JURISDICTION_ID,
+            CASE_TYPE_ID,
+            id
+        );
+
+        User defendant = userService.getUser(authorisation);
+        return caseIdsGivenUserIdHasAccessTo.stream()
+            .map((caseId) -> readCase(defendant, caseId))
```
jenkins-reform-hmcts: MINOR Remove the parentheses around the "caseId" parameter [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AS1611)

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -65,45 +76,125 @@ public CCDCaseApi(
...
+            authTokenGenerator.generate(),
+            anonymousCaseWorker.getUserDetails().getId(),
+            JURISDICTION_ID,
+            CASE_TYPE_ID,
+            id
+        );
+        if (letterHolderCases.size() > 1) {
+            throw new DefendantLinkingException("More than 1 case a letter holder ID has access to found");
+        }
+        if (letterHolderCases.size() == 0) {
```
jenkins-reform-hmcts: MINOR Use isEmpty() to check whether the collection is empty or not. [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AS1155)

src/main/java/uk/gov/hmcts/cmc/claimstore/services/UserService.java
```diff
@@ -21,6 +22,14 @@ public UserDetails getUserDetails(String authorisation) {
         return idamApi.retrieveUserDetails(authorisation);
     }
 
+    public User getUser(String authorisation) {
+        return new User(authorisation, idamApi.retrieveUserDetails(authorisation));
+    }
+
+    public User authenticateAnonymousCaseWorker() {
+        return new User("jkljasd", new UserDetails("1", "1@1.com", "1", "123"));
```
rlewan: Do we want those values?

timja-kainos: oh right, I forgot to do this :D :D

src/main/java/uk/gov/hmcts/cmc/claimstore/services/search/DBCaseRepository.java
```diff
@@ -38,13 +39,26 @@ public DBCaseRepository(
...
-        }
+    public Claim linkDefendant(String externalId, String defendantId, String authorisation) {
+        String notFoundErrorMessage = "Claim not found by external id: " + externalId;
+        
+        Claim claim = claimRepository.getClaimByExternalId(externalId)
+            .orElseThrow(() -> new NotFoundException(notFoundErrorMessage));
+        claimRepository.linkDefendant(claim.getId(), defendantId);
+
+        claim = claimRepository.getClaimByExternalId(externalId)
+            .orElseThrow(() -> new NotFoundException(notFoundErrorMessage));;
```
jenkins-reform-hmcts: MINOR Remove this empty statement. [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AEmptyStatementUsageCheck)

PR #176, opened 2018-02-02
ROC-3026 Create ccd response domains

dharmendrak:


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/response/ResponseMapper.java
```diff
@@ -0,0 +1,59 @@
...
+        this.statementOfTruthMapper = statementOfTruthMapper;
+        this.partyMapper = partyMapper;
+    }
+
+    @Override
+    public CCDResponse to(FullDefenceResponse response) {
+        CCDResponse.CCDResponseBuilder builder = CCDResponse.builder();
+
+        response.getFreeMediation()
+            .ifPresent(freeMediation -> builder.freeMediation(freeMediation.name().toLowerCase()));
```
sabahirfan: do we need to change the case here?

ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/response/ResponseMapper.java
```diff
@@ -0,0 +1,59 @@
...
+        builder.defence(response.getDefence());
+
+        return builder.build();
+    }
+
+    @Override
+    public FullDefenceResponse from(CCDResponse response) {
+
+        return new FullDefenceResponse(
+            Response.FreeMediationOption.valueOf(response.getFreeMediation().toUpperCase()),
```
sabahirfan: optional but may write utility method to covert from case insensitive name if we require the same at most places.
```
public static <T extends Enum<?>> T toEnum(Class<T> enumeration, String name) {
    for (T each : enumeration.getEnumConstants()) {
        if (each.name().compareToIgnoreCase(name) == 0) {
            return each;
        }
    }
    return null;
}
 ```

PR #182, opened 2018-02-06
ROC-3027: Fix save claim to CCD 

sabahirfan:
[ROC-3027](https://tools.hmcts.net/jira/browse/ROC-3027)

### Change description ###
This PR fixes the issue occurred during merge. Fix generates the correct JSON data.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/CoreCaseDataService.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/CoreCaseDataService.java
@@ -1,7 +1,5 @@
 package uk.gov.hmcts.cmc.claimstore.services.ccd;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/CoreCaseDataService.java
```diff
@@ -62,7 +66,7 @@ public CaseDetails save(String authorisation, Claim claim) {
                 .save(
                     authorisation,
                     eventRequestData,
-                    jsonMapper.toJson(ccdCase),
+                    toJson(ccdCase),
```
rlewan: Why can't we use `JsonMapper` @sabahirfan ?

sabahirfan: `jsonMapper` returns String object and if we pass `String` to `CCD Client` we are getting `400` Bad Request.

rlewan: Have you investigated why there's a 400 response? It should work with the `String`.

sabahirfan: Looks like it was due to double serialisation causing this by generating escaped characters.

PR #190, opened 2018-02-09
Remove JWTHelper

timja-kainos:


src/aat/java/uk/gov/hmcts/cmc/claimstore/tests/Bootstrap.java
```diff
@@ -57,4 +55,9 @@ public String getUserId() {
         return userId;
     }
 
+    private String decodeUserId(String authorisation) {
+        DecodedJWT decodedToken = JWT.decode(authorisation);
+        return decodedToken.getClaims().get("id").asString();
```
blucas: You shouldn't be looking at **any** of the claims in the JWT.

PR #197, opened 2018-02-14
Suppress dependency check warnings

timja-kainos:
Removed old warnings and added new suppressions that don't apply,
have added notes next to each one.


dependency-check-suppressions.xml
```diff
@@ -1,59 +1,66 @@
...
+    <suppress>
+        <notes><![CDATA[
+			something to do with oracle, doesn't look like it applies file name: jersey-apache-client4-1.19.1.jar
+			]]></notes>
+        <gav regex="true">^com\.sun\.jersey\.contribs:jersey-apache-client4:.*$</gav>
+        <cve>CVE-2006-0550</cve>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[
+			only applies when tomcat is a service, file name: tomcat-annotations-api-8.5.27.jar
```
rlewan: What do you mean by "tomcat is a service"? Boot uses Tomcat internally.

timja-kainos: is run as a service

PR #200, opened 2018-02-20
ROC-3142: save claim in ccd only

sabahirfan:
[ROC-3142](https://tools.hmcts.net/jira/browse/ROC-3142)

### Change description ###
This PR updated the CCD claim save feature, so that it will be either stored in claim-store db only or in CCD based on whether CCD is enabled or not. Currently we are storing in both repositories. 

- [ ] Yes
- [x] No


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
```diff
@@ -17,6 +18,7 @@
 
 // Create these fields in JSON when serialize Java object, ignore them when deserialize.
 @JsonIgnoreProperties(value = {"totalAmountTillToday", "totalAmountTillDateOfIssue"}, allowGetters = true)
+@Builder
```
timja-kainos: not sure how I feel about lombok being added here =/

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ReferenceNumberService.java
```diff
@@ -0,0 +1,27 @@
...
+public class ReferenceNumberService {
+
+    private final ReferenceNumberRepository referenceNumberRepository;
+
+    @Autowired
+    public ReferenceNumberService(ReferenceNumberRepository referenceNumberRepository) {
+        this.referenceNumberRepository = referenceNumberRepository;
+    }
+
+    public String getReferenceNumber(Boolean claimantRepresented) {
```
timja-kainos: use the primitive boolean? so it can't be null?

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
timja-kainos: @rlewan @geek-soft

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ReferenceNumberService.java
sabahirfan: done

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/CoreCaseDataService.java
```diff
@@ -41,6 +44,8 @@
     private final ResponseMapper responseMapper;
     private final SettlementMapper settlementMapper;
     private final UserService userService;
+    private final JsonMapper jsonMapper;
+    private final ReferenceNumberService referenceNumberService;
 
     @Autowired
     public CoreCaseDataService(
```
jenkins-reform-hmcts: MAJOR Constructor has 9 parameters, which is greater than 7 authorized. [rule](https://sonar.reform.hmcts.net/coding_rules#rule_key=squid%3AS00107)

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/ReferenceNumberRepository.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.hmcts.cmc.claimstore.repositories;
+
+import org.skife.jdbi.v2.sqlobject.SqlQuery;
+
+public interface ReferenceNumberRepository {
+
+    @SqlQuery("SELECT next_reference_number()")
+    String getReferenceNoForCitizen();
```
rlewan: Can you please change `No` to `Number`? I found it slightly confusing at a first glance.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/ReferenceNumberRepository.java
```diff
@@ -0,0 +1,13 @@
...
+
+    @SqlQuery("SELECT next_legal_rep_reference_number()")
+    String getReferenceNoForLegal();
```
rlewan: Same here.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
rlewan: I would prefer to not add Lombok if it's not really needed. `Claim` constructor is quite big but we probably need to invoke it in only one place in the code.

src/test/java/uk/gov/hmcts/cmc/claimstore/services/ReferenceNumberServiceTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Test
+    public void shouldGetReferenceNumberForCitizen() {
+
+        when(referenceNumberRepository.getReferenceNoForCitizen()).thenReturn(CITIZEN_REFERENCE);
+
+        Boolean claimRepresented = false;
+        String referenceNumber = referenceNumberService.getReferenceNumber(claimRepresented);
+
+        //then
```
rlewan: If you are using BDD-style comments then please use the full structure, i.e.
```
// given 
...
// when
...
// then
...
```
or remove them completely. Consider [`BDDMockito`](http://static.javadoc.io/org.mockito/mockito-core/2.15.0/org/mockito/BDDMockito.html) as well.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
sabahirfan: In my opinion, the question is not `lombok`, but do we need a `Builder` for this class? It shouldn't matter whether we choose over the shelf or write our own `Builder`. 
Whether it is invoked from single place or multiple, it leaves smell in the code.I would prefer not to call a constructor of class where I have to pass null for more than half of the parameters. Please advise if I am missing anything.

rlewan: Good point @sabahirfan. In that case I would say I'm ok with having the builder, my worry was that adding Lombok would incentivise people to make `Claim` a mutable object (I hope it doesn't generate any setters in this setup).

sabahirfan: No it's not generating setters just the Builder.

PR #203, opened 2018-02-20
Roc-3031: Handle settlement reached at for Counter sign

dharmendrak:


src/main/java/uk/gov/hmcts/cmc/claimstore/services/OffersService.java
```diff
@@ -31,53 +33,67 @@ public OffersService(
...
         assertSettlementIsNotReached(claim);
 
         Settlement settlement = claim.getSettlement()
-            .orElseThrow(() -> new ConflictException("Offer has not been made yet."));
+            .orElseThrow(conflictOfferIsNotMade());
         settlement.countersign(party);
 
-        caseRepository.reachSettlementAgreement(claim, settlement, authorisation,
-            userAcion("OFFER_COUNTERSIGNED_BY", party.name()));
+        caseRepository.reachSettlementAgreement(claim, settlement, authorisation, "SETTLED_PRE_JUDGMENT");
```
rlewan: Enum constant `.name()` can be used here.

dharmendrak: used, thanks

PR #219, opened 2018-02-26
ROC-2832: add unit tests

yarpo:
https://tools.hmcts.net/jira/browse/ROC-2832

### Change description ###
Added unit tests

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/cmc/claimstore/services/ccd/CoreCaseDataServiceTest.java
```diff
@@ -0,0 +1,222 @@
+package uk.gov.hmcts.cmc.claimstore.services.ccd;
```
timja-kainos: These tests seem very fake, we shouldn't be verifying that ccd is called when we call all the api methods.
I would recommend keeping the exception test, and making sure the saves are covered by api and functional tests

I know they're being written to improve coverage to get the build passing but I don't like this style of testing by verifying methods are called

PR #223, opened 2018-02-28
ROC-2832: add migration config to Jenkinsfile_CNP

yarpo:
https://tools.hmcts.net/jira/browse/ROC-2832

### Change description ###
Updated Jenkinsfile_CNP. 

```
[x] Yes
[ ] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -15,7 +15,7 @@ List<LinkedHashMap<String, Object>> secrets = [
   secret('IdamSystemUpdateUserUsername', 'IDAM_SYSTEM_UPDATE_USER_USERNAME'),
   secret('IdamSystemUpdateUserPassword', 'IDAM_SYSTEM_UPDATE_USER_PASSWORD'),
   secret('IdamAnonymousCaseworkerUsername', 'IDAM_ANONYMOUS_CASEWORKER_USERNAME'),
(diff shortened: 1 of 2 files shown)


Jenkinsfile_CNP
```diff
@@ -9,7 +9,13 @@ List<LinkedHashMap<String, Object>> secrets = [
   secret('AatTestUserUsername', 'SMOKE_TEST_CITIZEN_USERNAME'),
   secret('AatTestUserUsername', 'SMOKE_TEST_SOLICITOR_USERNAME'),
   secret('AatTestUserPassword', 'SMOKE_TEST_USER_PASSWORD'),
-  secret('AatTestUserEmailPattern', 'GENERATED_USER_EMAIL_PATTERN')
+  secret('AatTestUserEmailPattern', 'GENERATED_USER_EMAIL_PATTERN'),
+  secret('ClaimStoreDbPassword', 'CLAIM_STORE_DB_PASSWORD'),
+  secret('ClaimStoreDbHost', 'CLAIM_STORE_DB_HOST'),
```
geek-soft: We do not have to retrieve HOST name from vault.

yarpo: it was easier way as host for different env are not the same:
prod-data-lb.moneyclaim.reform.hmcts.net
test-data-lb.moneyclaim.reform.hmcts.net

yarpo: unless we only want to execute migration on `after('deploy:prod')` and we don't care for aat? [just to make sure it works well before production data]

PR #248, opened 2018-03-13
ROC-3252 Add fallback for bulk print

dharmendrak:
https://dev-build.reform.hmcts.net/blue/organizations/jenkins/hmcts%2Fcmc-claim-store/detail/PR-248/26/pipeline/


src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
```diff
@@ -0,0 +1,100 @@
...
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@TestPropertySource(
+    properties = {
+        "document_management.api_gateway.url=false",
+        "core_case_data.api.url=false",
+        "send-letter.url=http://localhost:8089/send"
+    }
+)
+@AutoConfigureWireMock(port = 8089)
```
rlewan: Please don't use a fixed port, there is no guarantee this port will be available on the Jenkins machine. Wiremock should provide a feature to determine an open port by itself.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
```diff
@@ -0,0 +1,100 @@
...
+                .withStatus(HttpStatus.BAD_REQUEST.value())
+                .withHeader("Content-Type", "application/json")
+            )
+        );
+
+        MvcResult result = makeRequest(SampleClaimData.submittedByClaimant())
+            .andExpect(status().isOk())
+            .andReturn();
+
+        verify(bulkPrintNotificationService, atLeast(1))
```
rlewan: Why do we use `atLeast` one here? Shouldn't there be only one notification expected?

rlewan: Or 3 as the test name suggests?

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
```diff
@@ -0,0 +1,100 @@
...
+            .willReturn(aResponse()
+                .withStatus(HttpStatus.INTERNAL_SERVER_ERROR.value())
+                .withHeader("Content-Type", "application/json")
+                .withBody("Internal server error occurred")));
+
+        MvcResult result = makeRequest(SampleClaimData.submittedByClaimant())
+            .andExpect(status().isOk())
+            .andReturn();
+
+        verify(bulkPrintNotificationService, atLeast(1))
```
rlewan: Same here.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/services/staff/BulkPrintStaffNotificationServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+            .builder()
+            .build();
+
+        when(pdfServiceClient.generateFromHtml(any(byte[].class), anyMap()))
+            .thenReturn(PDF_CONTENT);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void shouldThrowNullPointerWhenGivenNullClaim() {
+        service.notifyFailedBulkPrint(null, null);
```
rlewan: There are 2 `null` values passed in here but the test mentions claim only. Please provide a value for the other one and an additional test to check that NPE is thrown when the other one is `null`.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/services/staff/BulkPrintStaffNotificationServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+        verify(emailService).sendEmail(senderArgument.capture(), emailDataArgument.capture());
+
+        EmailAttachment emailAttachment = emailDataArgument.getValue()
+            .getAttachments()
+            .get(0);
+
+        String expectedPinLetterFileName
+            = DocumentNameUtils.buildDefendantLetterFileBaseName(claim.getReferenceNumber());
+
+        assertThat(emailAttachment.getContentType()).isEqualTo("application/pdf");
```
rlewan: Can you please check if Spring provides a constant for this content type?

src/main/java/uk/gov/hmcts/cmc/claimstore/services/staff/BulkPrintStaffNotificationService.java
```diff
@@ -0,0 +1,81 @@
...
+    public void notifyFailedBulkPrint(List<Document> documents, Claim claim) {
+        EmailContent emailContent = emailContentProvider.createContent(wrapInMap(claim));
+
+        EmailAttachment defendantLetter = pdf(
+            createPdf(documents.get(0)),
+            buildDefendantLetterFileBaseName(claim.getReferenceNumber())
+        );
+
+        EmailAttachment sealedClaim = pdf(
+            createPdf(documents.get(1)),
```
rlewan: Can you please check if we have an automatic test which verifies that the documents will always appear in this order?

dharmendrak: there is an unit test

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
```diff
@@ -0,0 +1,108 @@
...
+        MvcResult result = makeRequest(SampleClaimData.submittedByClaimant())
+            .andExpect(status().isOk())
+            .andReturn();
+
+        verify(bulkPrintNotificationService, never())
+            .notifyFailedBulkPrint(any(List.class), eq(deserializeObjectFrom(result, Claim.class)));
+    }
+
+    @Test
+    public void shouldTrySendingLetterThreeTimesOnFailuresWhenBulkPrintFailsWithSendLetterException() throws Exception {
```
geek-soft: Please rename, test does not check retry feature at all, just handling failure.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
```diff
@@ -0,0 +1,108 @@
...
+        MvcResult result = makeRequest(SampleClaimData.submittedByClaimant())
+            .andExpect(status().isOk())
+            .andReturn();
+
+        verify(bulkPrintNotificationService)
+            .notifyFailedBulkPrint(any(List.class), eq(deserializeObjectFrom(result, Claim.class)));
+    }
+
+    @Test
+    public void shouldTrySendingLetterThreeTimesOnFailuresWhenBulkPrintFailsWithRestClientException() throws Exception {
```
geek-soft: The same here please

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+        when(authTokenGenerator.generate()).thenReturn(AUTHORISATION_TOKEN);
+
```
geek-soft: Remove blank line for consistency.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
```diff
@@ -0,0 +1,108 @@
...
+        wireMockServer.stubFor(post(urlEqualTo("/send/letters"))
+            .willReturn(aResponse()
+                .withStatus(HttpStatus.INTERNAL_SERVER_ERROR.value())
```
geek-soft: Can you explain the difference in the way application handles 400 response and 500 response? I am trying to see why we have two test cases simulating these two response code.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/services/staff/BulkPrintStaffNotificationServiceTest.java
```diff
@@ -0,0 +1,120 @@
...
+    public void shouldSendEmailWithExpectedPDFAttachments() throws IOException {
+        service.notifyFailedBulkPrint(documents, claim);
+
+        verify(emailService).sendEmail(senderArgument.capture(), emailDataArgument.capture());
+
+        EmailAttachment emailAttachment = emailDataArgument.getValue()
+            .getAttachments()
+            .get(0);
+
+        String expectedPinLetterFileName
```
geek-soft: That test does not guarantee that in real life first document will be a PIN letter and second one will be sealed claim copy. 

I can change `publisher.publishEvent(new DocumentReadyToPrintEvent(event.getClaim(), defendantLetterDocument , sealedClaimDocument));` to be` publisher.publishEvent(new DocumentReadyToPrintEvent(event.getClaim(), sealedClaimDocument, defendantLetterDocument));` and tests will keep passing while in application will start sending PIN letters as sealed claims and vice versa.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/staff/BulkPrintStaffNotificationService.java
geek-soft: The test is useless in that matter, refactoring can easily bring breaking change that is not detected. See my comment above.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/services/staff/BulkPrintStaffNotificationServiceTest.java
dharmendrak: BulkPrintServiceTest checks for the order of documents and will fail if order changes

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/MockSpringTest.java
```diff
@@ -23,11 +23,10 @@
 import uk.gov.hmcts.reform.document.DocumentMetadataDownloadClientApi;
 import uk.gov.hmcts.reform.document.DocumentUploadClientApi;
 import uk.gov.hmcts.reform.pdf.service.client.PDFServiceClient;
-import uk.gov.hmcts.reform.sendletter.api.SendLetterApi;
 import uk.gov.service.notify.NotificationClient;
 
 @RunWith(SpringRunner.class)
-@SpringBootTest
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
```
rlewan: I think that's not needed, Spring does not need to open a port, only Wiremock does.

dharmendrak: you are right, it works fine without it, however https://cloud.spring.io/spring-cloud-contract/1.1.x/multi/multi__spring_cloud_contract_wiremock.html does say to use it.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
dharmendrak: it has no purpose as it is handling both exceptions in same way. However wanted to keep two exceptions in sendLetterApi, if other user want to handle them separately.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
dharmendrak: done

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
dharmendrak: done

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/BulkPrintRequestTest.java
dharmendrak: done

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/services/staff/BulkPrintStaffNotificationServiceTest.java
geek-soft: No it does not, it checks only that the order is not changed within the `BulkPrintServiceTest`

geek-soft:
@dharmendrak please change `DocumentReadyToPrintEvent` to have two properties (defendantLetterDocument and sealedClaimDocument) instead of using `List<Document> documents`. That will guarantee ordering. Then in unit tests make sure letter goes first followed by sealed claim.

PR #262, opened 2018-03-16
Disable test that only runs against CCD

timja-kainos:


timja-kainos:
Merging to fix build and unblock people

PR #263, opened 2018-03-17
Remove 404 from root

timja-kainos:
Application insights was registering lots of failures because of the
azure keep alive which hits the app every 5 seconds on / (its not
configurable)

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/RootController.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/RootController.java
@@ -4,9 +4,6 @@ import org.springframework.http.MediaType;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RestController;
 
-import java.util.Collections;
-import java.util.Map;
-
 @RestController
 public class RootController {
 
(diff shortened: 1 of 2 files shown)


timja-kainos:
Build is green ignore sandbox build error:
https://dev-build.reform.hmcts.net/job/hmcts/job/cmc-claim-store/view/change-requests/job/PR-263/

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/RootController.java
```diff
@@ -0,0 +1,22 @@
...
+public class RootController {
+
+    /*
+     * Azure hits us on / every 5 seconds to prevent it sleeping the application
+     * Application insights registers that as a 404 and adds it as an exception,
+     * This is here to reduce the noise
+     */
+    @GetMapping(value = "/", consumes = MediaType.ALL_VALUE, produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public Map root() {
+        return Collections.singletonMap("hello", "world!");
```
geek-soft: I would not expect to see this in production grade application. I would expect either blank page / empty JSON object or page / JSON object with some explanation why this path exists.

timja-kainos: @geek-soft removed

PR #269, opened 2018-03-20
Fix error when evidence is not provided

sabahirfan:
This PR fixes the below error in case no evidence was provided.

```
claim-store-api_1                      | 2018-03-20T13:45:22.921+0000 ERROR [http-nio-4400-exec-2] uk.gov.hmcts.cmc.claimstore.controllers.advices.ResourceExceptionHandler:41: null
claim-store-api_1                      |     at uk.gov.hmcts.cmc.claimstore.documents.ClaimDataContentProvider.createContent(ClaimDataContentProvider.java:72)
claim-store-api_1                      |     at uk.gov.hmcts.cmc.claimstore.documents.ClaimContentProvider.createContent(ClaimContentProvider.java:58)
claim-store-api_1
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/documents/ClaimDataContentProvider.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/documents/ClaimDataContentProvider.java
@@ -16,6 +16,7 @@ import uk.gov.hmcts.cmc.domain.models.legalrep.StatementOfTruth;
 
 import java.math.BigDecimal;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -68,8 +69,10 @@ public class ClaimDataContentProvider {
         List<EvidenceContent> evidences = null;
         Optional<Evidence> evidence = claim.getClaimData().getEvidence();
         if (evidence.isPresent()) {
-            evidences = evidence.get().getRows()
+            evidences = Optional.ofNullable(evidence.get().getRows())
+                .orElseGet(Collections::emptyList)
                 .stream()
+                .filter(Objects::nonNull)
                 .map(e -> new EvidenceContent(e.getType().getDescription(), e.getDescription().orElse(null)))
                 .collect(Collectors.toList());
         }


src/main/java/uk/gov/hmcts/cmc/claimstore/documents/ClaimDataContentProvider.java
```diff
@@ -68,8 +69,10 @@ public ClaimContent createContent(Claim claim) {
         List<EvidenceContent> evidences = null;
         Optional<Evidence> evidence = claim.getClaimData().getEvidence();
         if (evidence.isPresent()) {
-            evidences = evidence.get().getRows()
+            evidences = Optional.ofNullable(evidence.get().getRows())
```
timja-kainos: why would evidence be present but not the rows?

sabahirfan: front-end is sending in case of no evidence was entered.
```
	"evidence": {
		"rows": []
	},
```

timja-kainos: that's an empty list then? and should be fine?

rlewan: Tim is correct -- please investigate the problem more throughly so that we know why this issue is happening.

geek-soft: Can we please investigate that more, we should avoid coding against empty lists.

Also I would say evidence object should be submitted only when list is populated. Can we fix frontend conversion and put some validation against empty list of rows?

sabahirfan: When this is mapped to Model object Jackson converts to null list/array if empty.

sabahirfan: This is what we get saved in claim-store database,
```
 "evidence": {},
```

timja-kainos: that doesn't seem right, fix might be be to stop sending that, a data migration to delete any of those should be safe too

geek-soft: I would be happy to accept current solution as technical debt however I had a chat with @sabahirfan and he is happy to make changes to:

1. send evidence object only when at least one row is populated
2. validate evidence object on backend side to make sure if submitted there is at least one row
3. check whether we have any of evidence data in production DB and migrate if necessary

sabahirfan: On point 2. There is one concern. We will have different validation logics for front-end and back-end.

PR #276, opened 2018-03-21
ROC-3370: Point service to new GOV.UK pages

yarpo:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/ROC-3370

### Change description ###
Update url to frontend in prod config

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #285, opened 2018-03-23
ROC-3370: Make first contact URL configurable

yarpo:
https://tools.hmcts.net/jira/browse/ROC-3370

### Change description ###
Added new var. For now it has the same value it used to have. In next PR ( https://github.com/hmcts/cmc-claim-store/pull/276 ) 

```
[ ] Yes
[x] No
```


src/main/resources/application.yml
```diff
@@ -63,10 +63,11 @@ dateCalculations:
 notifications:
   govNotifyApiKey: ${GOV_NOTIFY_API_KEY:}
   frontendBaseUrl: "${FRONTEND_BASE_URL:}"
+  respondToClaimUrl: "${RESPOND_TO_CLAIM_URL:}"
   templates:
     email:
       claimantClaimIssued: "58e645b8-bcd6-4462-822c-f17da9dd3dc7"
-      defendantClaimIssued: "c38550a0-a981-4698-a644-481ad3dd8228"
+      defendantClaimIssued: "49027a97-412e-45ee-86df-bf2fa6df6d8c"
```
timja-kainos: I would have thought we would use the existing one and just update the parameter on the day?

geek-soft: Had that idea as well but wanted have deployment that does not require human intervention. 

With web hooks in place we would need to be doing it this anyway.

Other thing is that in the future I would like to have all URLs configurable via application config, no updates on GOV.UK Notify, mainly to have single place to change.

PR #289, opened 2018-03-25
Calculate interest to issue date on calculation service

timja-kainos:


timja-kainos:
Merging to prevent conflicts in new fix

PR #298, opened 2018-04-03
ROC-3481: on PDFs total amount should not reflect interests

yarpo:
https://tools.hmcts.net/jira/browse/ROC-3481

### Change description ###
On PDF documents we don't want to display total amount with interest. 
On page interests should be added.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/documents/ClaimDataContentProvider.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/documents/ClaimDataContentProvider.java
@@ -56,7 +56,6 @@ public class ClaimDataContentProvider {
                 claim.getIssuedOn(),
                 claim.getIssuedOn()
             );
-            totalAmountComponents.add(interestContent.getAmountRealValue());
         }
 
         Optional<StatementOfTruth> optionalStatementOfTruth = claim.getClaimData().getStatementOfTruth();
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/cmc/claimstore/documents/ClaimDataContentProvider.java
```diff
@@ -56,7 +56,6 @@ public ClaimContent createContent(Claim claim) {
                 claim.getIssuedOn(),
                 claim.getIssuedOn()
             );
-            totalAmountComponents.add(interestContent.getAmountRealValue());
```
geek-soft: That change impacts sealed claim copy. @yarpo please make sure business (Art and Mark) understand that.

IMO `Claim amount: / Total:` without interest is less readable especially that section contains interest information.

Maybe more appropriate solution would be publishing PDF's into document store.

yarpo: I believe publishing would be an interesting solution. 

I agree that "total amount" which is not "so total" is misleading.

sabahirfan: @yarpo I think we still need to add interest in sealed claim, but when the copy is downloaded it should not add daily interest  based on the days passed. In short no matter when user downloads the pdf copy it should reflect the same amounts as when sealed claim was generated. 

Agree with @geek-soft If we have `document-store` enabled/working probably better to store sealed claim there, and later retrieve it from there for copy.

yarpo:
New PR:
https://github.com/hmcts/cmc-claim-store/pull/303

src/main/java/uk/gov/hmcts/cmc/claimstore/documents/ClaimDataContentProvider.java
yarpo: Please see:
https://github.com/hmcts/cmc-claim-store/pull/303

PR #299, opened 2018-04-04
ROC-3460: One app insights per product

rlewan:

https://tools.hmcts.net/jira/browse/ROC-3460

### Change description ###

A change to start using a single App Insights insights instance which aggregates all services per environment.


Jenkinsfile_parameterized
```diff
@@ -1,5 +1,5 @@
 #!groovy
-@Library("Infrastructure") _
+@Library("Infrastructure@feature/ROC-3460-One-app-insights-per-product") _
```
rlewan: Will update after the referenced part is merged.

infrastructure/main.tf
```diff
@@ -46,12 +46,13 @@ data "vault_generic_secret" "system_update_password" {
 }
 
 module "claim-store-api" {
-  source = "git@github.com:contino/moj-module-webapp.git"
+  source = "git@github.com:contino/moj-module-webapp.git?ref=feature/ROC-3460-One-app-insights-per-product"
```
rlewan: Will update after the referenced part is merged.

PR #301, opened 2018-04-06
ROC-3407: Make defence text optional

geek-soft:

https://tools.hmcts.net/jira/browse/ROC-3407

### Change description ###

This change removes 'NotBlank' validation constraint from defence text field to allow state paid defences without defence text. Information of when and how defendant paid should be enough in the court.

```
[ ] Yes
[x] No
```


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/FullDefenceResponse.java
```diff
@@ -70,8 +68,8 @@ public DefenceType getDefenceType() {
         return defenceType;
     }
 
-    public String getDefence() {
-        return defence;
+    public Optional<String> getDefence() {
```
dharmendrak: didn't get why optional when minimum size is 1

geek-soft: Minimal size is 1 when defence is submitted to prevent accepting `""`. Optional is needed because I removed constraint that required the value so now submitting `null` or `undefined` is possible.

PR #303, opened 2018-04-06
ROC-3481: fix bug

yarpo:
https://tools.hmcts.net/jira/browse/ROC-3481

### Change description ###
On sealed claim and claimant copy pdf total amount should always be the same.

```
[ ] Yes
[x] No
```


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/InterestDate.java
```diff
@@ -60,6 +60,21 @@ public InterestEndDateType getEndDateType() {
         return endDateType;
     }
 
+    @JsonIgnore
```
timja-kainos: hmm?

src/main/java/uk/gov/hmcts/cmc/claimstore/documents/ClaimDataContentProvider.java
```diff
@@ -97,12 +99,15 @@ public ClaimContent createContent(Claim claim) {
...
-            amountBreakDown.getRows()
-                .stream()
-            .filter(row -> row != null && row.getAmount() != null)
-            .map(AmountRowContent::new)
-            .collect(Collectors.toList())
+            removeEmptyRows(amountBreakDown.getRows())
         );
     }
 
+    private static List<AmountRowContent> removeEmptyRows(List<AmountRow> rows) {
```
timja-kainos: it does more than remove empty rows, it converts to a format for display

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/InterestDate.java
yarpo: We don't want it to be mapped into JSON.

src/main/java/uk/gov/hmcts/cmc/claimstore/documents/ClaimDataContentProvider.java
```diff
@@ -97,12 +99,15 @@ public ClaimContent createContent(Claim claim) {
...
-            .map(AmountRowContent::new)
-            .collect(Collectors.toList())
+            mapToAmountRowContent(amountBreakDown.getRows())
         );
     }
 
+    private static List<AmountRowContent> mapToAmountRowContent(List<AmountRow> rows) {
+        return rows
+            .stream()
+            .filter(row -> row != null && row.getAmount() != null)
```
sabahirfan: You may use  streams filter here. 'filter(Objects::nonNull)'

PR #304, opened 2018-04-09
ROC-3461 Track dependency calls, add events

timja-kainos:
https://tools.hmcts.net/jira/browse/ROC-3461
https://tools.hmcts.net/jira/browse/ROC-3463

### Change description ###
Java applications need the appinsights agent present in the deployment so that instrumentation can be done.
Automatic instrumentation is done for:

* HttpUrlConnection (Not https 😢)
* ApacheHttpClient 4.3.x
* Jedis (Java redis client)
* Prepared statements JDBC, (Doesn't seem to be working for the claim-store's JDBI implementation)
Instrumentation can be easily configured by providing an AI-Agent.xml file
See:
https://github.com/hmcts/cmc-claim-store/blob/7e5ba9a5ae93f6122980e360ecb87ffda1a30444/AI-Agent.xml

Application map:
image
Localhost will be split out after everything is on different hosts

I also added event tracking while exploring how application insights works, and it looks really useful for analytics, azure dashboard and hopefully troubleshooting and alerting.

Note this requires:
https://github.com/hmcts/java-logging/pull/57
and
https://github.com/hmcts/moj-jenkins-library/pull/197

```
[ ] Yes
[X] No
```


build.gradle
```diff
@@ -134,8 +134,6 @@ dependencies {
 
   compile group: 'uk.gov.service.notify', name: 'notifications-java-client', version: '3.8.0-RELEASE'
 
-  compile group: 'uk.gov.hmcts.reform', name: 'java-logging-spring', version: versions.reformJavaLogging
-  compile group: 'uk.gov.hmcts.reform', name: 'java-logging-httpcomponents', version: versions.reformJavaLogging
```
doncem: by removing these components, `java-logging` one is not included. if that's intentional - all good

web.config
```diff
@@ -7,7 +7,7 @@
     <httpPlatform stdoutLogEnabled="true"
                   stdoutLogFile="%HOME%\LogFiles\application\stdout"
                   processPath="%JAVA_HOME%\bin\java.exe"
-        arguments="-Djava.net.preferIPv4Stack=true -Dserver.port=%HTTP_PLATFORM_PORT% -jar &quot;%HOME%\site\wwwroot\claim-store.jar&quot;">
+        arguments="-javaagent:&quot;%HOME%\site\wwwroot\applicationinsights-agent-2.0.2.jar&quot; -Djava.net.preferIPv4Stack=true -Dserver.port=%HTTP_PLATFORM_PORT% -jar &quot;%HOME%\site\wwwroot\claim-store.jar&quot;">
```
doncem: Ah that is what's missing from agent additional config 👏 . Can we have an update in `java-logging-appinsights` readme?

By the way, agent jar comes together from library, is it not visible / accessible just from there? Then could get rid of extra jar in project root lib folder

timja-kainos: Not without unzipping it and extracting the jar out, I'll remove it from java-logging after I've got this to prod.
Will be updating rhubarb and the onboarding docs as well as java-logging

build.gradle
timja-kainos: I compared the logging output and it looked the same to me, 
Wasn't really sure if I should add the `java-logging` dep here.
To be honest java-logging isn't doing much now, for us its just providing the app insights filters etc which won't be needed when the appinsights spring boot starter is released in the next month or two hopefully

timja-kainos: Turns out the dep was already there but it was version 1.4.0 coming from the pdf service client, I've added the dep explicitly here to keep them in sync

PR #320, opened 2018-04-17
ROC-3535 Publish to docker hub

timja-kainos:


Dockerfile
```diff
@@ -1,6 +1,12 @@
+FROM gradle:jdk8 as builder
```
geek-soft: Is there any reason we are changing base image and an approach to build image inside Docker?

timja-kainos: dockerhub will be automatically building the image, docker build needs to actually build stuff now.

timja-kainos: Another option I thought of was to provide 2 docker files
One for docker hub
and one for local build
It would mean we keep our speed for building images at the cost of maintaining two files

PR #325, opened 2018-04-25
ROC-3440: Automate pdf checking

varma02:
https://tools.hmcts.net/jira/browse/ROC-3440

### Change description ###
Automate PDF checking for fields in claim-store


src/aat/java/uk/gov/hmcts/cmc/claimstore/tests/functional/SealedClaimPdfTest.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Arrays;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SealedClaimPdfTest extends BaseTest {
+
+    private User claimant;
+    @Autowired
+    private ClaimIssueReceiptService claimIssueReceiptService;
+    static File fileName = new File("/Users/kiranv/cmc/claim-store/src/aat/java/uk/gov/hmcts/cmc/claimstore/tests/functional/999MC042-claim-form.pdf");
```
timja-kainos: to get the file need to download it from the document controller after the case has been created

geek-soft: Hmm, expected something a bit different when I had a chat with @varma02 the other day. I though the approach was to have the logic to assert PDF content first (as it is easier) with assertions agreed with QA and then change the source of the PDF to take it from REST calls.

timja-kainos: shouldn't be much harder? 
we also have a story in this sprint to download sealed claim from the frontend so its needed functionality too

open to either way but i thought the agreed approach was to download the PDF after saving the claim?

varma02: I had a chat with Vivek and we agreed to do asserts on Claim Ref, Issued on, Claimant name, Defendant name, Claim amount, Total amount, Respond before date and add other data later if needed.

build.gradle
```diff
@@ -155,6 +155,7 @@ dependencies {
   compile group: 'io.github.openfeign', name: 'feign-httpclient', version: '9.5.1'
 
   compile group: 'com.microsoft.azure', name: 'applicationinsights-logging-logback', version: '2.0.0-BETA'
+  compile group: 'org.pdfbox', name: 'com.springsource.org.pdfbox', version: '0.7.3'
```
timja-kainos: you don't want this all compile
I think `aatCompile` is what you want
otherwise you're bloating the deployed version with something it doesn't need

varma02:
some rework required for assertions but everything else is complete.

src/aat/java/uk/gov/hmcts/cmc/claimstore/tests/functional/SealedClaimPdfTest.java
```diff
@@ -0,0 +1,108 @@
...
+        assertThat(pdfAsText).contains("Issued on: " + Formatting.formatDate(createdCase.getIssuedOn()));
+        assertThat(pdfAsText).contains("Name: " + createdCase.getClaimData().getClaimant().getName());
+        assertThat(pdfAsText).contains("Address: "
+            + createdCase.getClaimData().getClaimant().getAddress().getLine1() + " \n"
+            + createdCase.getClaimData().getClaimant().getAddress().getLine2() + " \n"
+            + createdCase.getClaimData().getClaimant().getAddress().getLine3() + " \n"
+            + createdCase.getClaimData().getClaimant().getAddress().getCity() + " \n"
+            + createdCase.getClaimData().getClaimant().getAddress().getPostcode());
+        assertThat(pdfAsText).contains("Name: " + createdCase.getClaimData().getDefendant().getName());
+        assertThat(pdfAsText).contains("Address: "
```
sabahirfan: can we please refactor out address and reuse?

varma02: done in next commit

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentsService.java
```diff
@@ -5,6 +5,8 @@
 
     byte[] getLegalSealedClaim(String externalId, String authorisation);
 
+    byte[] getSealedClaim(String externalId, String authorisation);
```
sabahirfan: Do we need this new method? I think rather than having two different methods(`getLegalSealedClaim ` & `getSealedClaim `) here, we can have just one method probably called `getSealedClaim ` and inside it's implementation we can distinguish based on `claim.getClaimData().isClaimantRepresented()`.

We can rename `LegalSealedClaimPdfService` to `SealedClaimPdfService` which creates both PDFs. 
```
public byte[] createPdf(Claim claim) {
 if (claim.getClaimData().isClaimantRepresented()) {
 return pdfServiceClient.generateFromHtml(
            documentTemplates.getLegalSealedClaim(),
            legalSealedClaimContentProvider.createContent(claim)
        );
} else {
 return <CITIZEN_PDF_GENERATED>;
}
```

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/DocumentsController.java
```diff
@@ -57,7 +57,7 @@ public DocumentsController(DocumentsService documentsService) {
         @PathVariable("externalId") @NotBlank String externalId,
         @RequestHeader(HttpHeaders.AUTHORIZATION) String authorisation
     ) {
-        byte[] pdfDocument = documentsService.getLegalSealedClaim(externalId, authorisation);
+        byte[] pdfDocument = documentsService.getSealedClaim(externalId, authorisation);
```
geek-soft: That is breaking change. Legal sealed copy is different and should not get changed

PR #330, opened 2018-04-30
ROC-3593: Suppress Staff emails on claim submission

jacmurphy:

https://tools.hmcts.net/jira/browse/ROC-3593

### Change description ###
This WIP-PR will focus on getting the Claim Submission fully functional and tested. It will later be expanded to all staff notifications.

Add a feature toggle called emailToStaff. Initially it is defaulted to false.
Tests still need to be made to make sure conditions respond accordingly.


src/main/resources/application.yml
```diff
@@ -85,6 +85,7 @@ staff-notifications:
 
 feature_toggles:
   watermark_pdf: false
+  emailToStaff: fasle
```
geek-soft: There is no such thing as `fasle` ;)

geek-soft: Other thing is that notifications should be enabled by default.

jacmurphy: Good spot! 
I was testing that it was working properly and swapped it from true to see if it worked. It will be changed in the next commit

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/services/staff/ClaimIssuedStaffNotificationServiceTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@TestPropertySource(
+    properties = {
+        "feature_toggles.emailToStaff=true"
```
dharmendrak: as it is defaulted to true, no need to set this property, instead test for false value, confirming staff email is not send.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CounterSignOfferWithStaffNotificationDisabledTest.java
```diff
@@ -0,0 +1,69 @@
...
+public class CounterSignOfferWithStaffNotificationDisabledTest extends BaseSaveTest {
+    @MockBean
+    protected SendLetterApi sendLetterApi;
+
+    @Captor
+    private ArgumentCaptor<EmailData> emailDataArgument;
+
+    @Test
+    public void shouldNotSendStaffNotificationWhenCounterSignRequestSubmitted() throws Exception {
+        Claim claim = claimStore.saveClaim(SampleClaimData.builder().build(), "1", LocalDate.now());
```
yarpo: what is "1"?

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CounterSignOfferWithStaffNotificationDisabledTest.java
```diff
@@ -0,0 +1,69 @@
...
+        caseRepository.linkDefendantV1(claim.getExternalId(), DEFENDANT_ID, BEARER_TOKEN);
+        Response response = SampleResponse.validDefaults();
+
+        caseRepository.saveDefendantResponse(claim, "defendant@mmail.com", response, BEARER_TOKEN);
```
yarpo: don't we have DEFENDANT_EMAIL const in any SampleResponse / SampleParty?

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CounterSignOfferWithStaffNotificationDisabledTest.java
```diff
@@ -0,0 +1,69 @@
...
+        Settlement settlement = new Settlement();
+        settlement.makeOffer(SampleOffer.validDefaults(), MadeBy.DEFENDANT);
+        caseRepository.updateSettlement(claim, settlement, BEARER_TOKEN, "OFFER_MADE_BY_DEFENDANT");
+
+        settlement.accept(MadeBy.CLAIMANT);
+        caseRepository.updateSettlement(claim, settlement, BEARER_TOKEN, "OFFER_ACCEPTED_BY_CLAIMANT");
+
+        makeRequest(claim.getExternalId());
+
+        verify(emailService, never()).sendEmail(eq("sender@example.com"), emailDataArgument.capture());
```
yarpo: same as above :)

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/services/staff/ClaimIssuedStaffNotificationServiceTest.java
yarpo: I'd actually leave it this way to make it clear that we want top test different config value than in other tests.

What I'd suggest to add is Java doc above that would describe what is this test for.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/SupportController.java
```diff
@@ -30,6 +31,7 @@
 
 @RestController
 @RequestMapping("/support")
+@ConditionalOnProperty(prefix = "feature_toggles", name = "emailToStaff")
```
yarpo: are you sure you want to disable the whole controller when you don't want to send emails to staff?

It looks a bit too nuke for me :)

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CounterSignOfferWithStaffNotificationDisabledTest.java
jacmurphy: 1 is the submitterID

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CounterSignOfferWithStaffNotificationDisabledTest.java
jacmurphy: I dont think so. Do you mean it is a global const ?

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/SupportController.java
jacmurphy: I agree it does seem a bit overkill.
Im open to a better solution

dharmendrak: there is no solution, bean it contains get disabled. so it need to disabled otherwise will give spring autowiring error

PR #279, opened 2018-03-21
ROC-3304: Defence PDF content revision

singh-surjeet:
### JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/ROC-3304

### Change description ###
Using single defendant response PDF across all user claimant/defendant/staff. No seal, renamed and most of content revised. 

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[X ] No
```

PR #338, opened 2018-05-10
ROC-3706: Add amended flag when defendant address is updated

rattur:

https://tools.hmcts.net/jira/browse/ROC-3706

### Change description ###

Add text '(as amended)' under the address section when defendant address details updated

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/resources/citizen/templates/document/defendantResponseReceipt.html
+++ b/src/main/resources/citizen/templates/document/defendantResponseReceipt.html
@@ -134,6 +134,10 @@
       padding-right: 0px;
     }
 
+    .keep-formatting {
+      white-space: pre-line;
+    }
+
   </style>
 </head>
 <body>
@@ -190,7 +194,10 @@
       </tr>
       {% endif %}
       <tr>
-        <td>Address:</td>
+        <td class="keep-formatting">Address:
+          {% if defendant.addressAmended %}
+          (as amended)
+          {% endif %}</td>
         <td>
           {{ defendant.address.line1 }}
           <br/>


rattur:
screen shot 2018-05-10 at 11 50 03

src/main/resources/citizen/templates/document/defendantResponseReceipt.html
```diff
@@ -190,7 +194,10 @@
       </tr>
       {% endif %}
       <tr>
-        <td>Address:</td>
+        <td class="keep-formatting">Address:
+          {% if defendant.addressAmended %}
+          (as amended)
```
timja-kainos: is this where its supposed to go? will this break robotics?
we used to have it after this address first line I think

rattur: That's what Marc James and Art would like

timja-kainos: will it break robotics?

rattur: I'm not sure Tim, will need to check

geek-soft: @rattur I managed to find previous version. Here is what we had before https://github.com/hmcts/cmc-claim-store/pull/279/files#diff-5b3cba7f1066682c02d7992300655a03L175.

@singh-surjeet removed that feature on 22nd Match.

I wonder whether bringing previous implementation would make it easier for RPA team. I guess we need to wait for response from them.

rattur: This new version is preferred from the previous implementation, therefore, had conversations and sent emails with the new pdf attached to Clem and RPA team. Will wait for the outcome

geek-soft:
Solution seems valid. I put DO NOT MERGE prefix as we need to confirm with RPA team that change is not going to break automation.

geek-soft:
Solution deployed on RPA side, removing DO NOT MERGE prefix.

PR #341, opened 2018-05-10
BigFix: Fixed unhealthy claim-store container

jacmurphy:
### Change description ###
wget was trying to download 'health' from 'http://localhost:4400/health' and failing, leading to an unhealthy container.
By using the --spider command, we check for a 200 response instead. 

```
[ ] Yes
[x] No
```


geek-soft:
Nice fix @jacmurphy. Can you please apply the same fix to other Docker images?

PR #342, opened 2018-05-11
ROC-3429: Duplicate interestDate under interest

rlewan:

https://tools.hmcts.net/jira/browse/ROC-3429

### Change description ###

This is the first step to deliver this story. 

```
[ ] Yes
[X] No
```

### DIFF
--- a/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
+++ b/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
@@ -62,7 +62,6 @@ public class ClaimData {
     @Valid
     private final Interest interest;
 
-    // Validated by InterDependentFields
     private final InterestDate interestDate;
 
     @Valid
--- a/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
+++ b/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
@@ -151,7 +151,13 @@ public class ClaimData {
     }
 
     public InterestDate getInterestDate() {
-        return interestDate;
+        if (interestDate != null) {
+            return interestDate;
+        } else if (interest != null) {
+            return interest.getInterestDate();
+        } else {
+            return null;
+        }
     }
 
     @JsonIgnore
--- a/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Interest.java
+++ b/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Interest.java
@@ -73,6 +73,10 @@ public class Interest {
         return Optional.ofNullable(specificDailyAmount);
     }
 
+    public InterestDate getInterestDate() {
+        return interestDate;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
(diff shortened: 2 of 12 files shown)


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Interest.java
```diff
@@ -69,25 +73,30 @@ public String getReason() {
...
     }
 
+    public InterestDate getInterestDate() {
+        return interestDate;
+    }
+
     @Override
-    public boolean equals(Object other) {
-        if (this == other) {
+    public boolean equals(Object o) {
```
timja-kainos: is o going to pass checkstyle?

domain-model/src/test/java/uk/gov/hmcts/cmc/domain/models/ClaimDataDeprecableTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.junit.Test;
+import uk.gov.hmcts.cmc.domain.models.sampledata.SampleClaimData;
+import uk.gov.hmcts.cmc.domain.models.sampledata.SampleInterest;
+import uk.gov.hmcts.cmc.domain.models.sampledata.SampleInterestDate;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ClaimDataDeprecableTest {
+
```
timja-kainos: extra line here?

domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleInterest.java
```diff
@@ -71,13 +73,20 @@ public SampleInterest withSpecificDailyAmount(BigDecimal specificDailyAmount) {
...
-            rate,
-            reason,
-            specificDailyAmount
+                type,
+                interestBreakdown,
+                rate,
+                reason,
+                specificDailyAmount,
+                interestDate
+
```
timja-kainos: extra line here

rlewan:
We noticed some problems when running Codecept tests against the updated Claim Store, need to sort that out first.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
```diff
@@ -152,7 +151,13 @@ public Interest getInterest() {
     }
 
     public InterestDate getInterestDate() {
-        return interestDate;
+        if (interestDate != null) {
+            return interestDate;
+        } else if (interest != null) {
+            return interest.getInterestDate();
+        } else {
+            return null;
```
geek-soft: Is there an chance that interest date is null? If so shouldn't we use `Optional<InterestDate>`.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Interest.java
geek-soft: I would love to revert that rename.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
rlewan: It's true for the current design but I don't think it's worth doing this refactoring. Once we move the field this will no longer be the case.

PR #346, opened 2018-05-15
Removed outdated section in the README

ckalista:
The section in the README explaining how to connect to the database from within the claim-store is no longer relevant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,12 +94,6 @@ Database will get initiated when you run `docker-compose up` for the first time 
 
 You don't need to migrate database manually since migrations are executed every time `docker-compose up` is executed.
 
-When you need to make ad-hoc connect to the database you can execute following command:
-
-```
-docker-compose exec claim-store-api psql -U postgres -d claimstore
-```
-
 ### API documentation
 
 API documentation is provided with Swagger:
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ Database will get initiated when you run `docker-compose up` for the first time 
 
 You don't need to migrate database manually since migrations are executed every time `docker-compose up` is executed.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,12 +94,6 @@ Database will get initiated when you run `docker-compose up` for the first time
 
 You don't need to migrate database manually since migrations are executed every time `docker-compose up` is executed.
```
geek-soft: That one has a hatch as technically not every `docker-compose up` starts the application and triggers migration. Maybe we could also replace `docker-compose up` with something that says that it happens on application start time.

README.md
```diff
@@ -97,7 +97,7 @@ You don't need to migrate database manually since migrations are executed every
 When you need to make ad-hoc connect to the database you can execute following command:
 
 ```
-docker-compose exec claim-store-api psql -U postgres -d claimstore
+docker exec -it cmc-integration-tests_shared-database_1 psql -U postgres -d claimstore
```
geek-soft: The only problem with this is that we introduce dependency between repositories. If someone changed integration-tests to remove database or change service name it is hard to see that this README needs an update as well.

timja-kainos: This is what I do for ad-hoc connection:
```
docker-compose exec shared-database psql -U postgres -d claimstore
```
In the integration tests project

Although I have it configured in my intellij which is what I use 99% of the time, more if I ever need to connect to idam or pay etc

README.md
ckalista: Absolutely, I have changed the wording.

README.md
ckalista: Agreed, I have removed it from here and will update the README in integration-tests instead.

PR #349, opened 2018-05-16
ROC-3728 more time paper event

timja-kainos:
https://tools.hmcts.net/jira/browse/ROC-3728

### Change description ###
Adds support for more time requested paper event

also:
https://github.com/hmcts/cmc-integration-tests/pull/185/files

```
[ ] Yes
[ ] No
```


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CaseEvent.java
```diff
@@ -5,7 +5,7 @@
     SUBMIT_CLAIM("submitClaimEvent"),
     DEFAULT_CCJ_REQUESTED("DefaultCCJRequested"),
     DEFENCE_SUBMITTED("DefenceSubmitted"),
-    MORE_TIME_REQUESTED("MoreTimeRequested"),
+    MORE_TIME_REQUESTED_ONLINE("MoreTimeRequestedOnline"),
```
rlewan: Don't a paper event as well?

timja-kainos: not really currently, as we aren't sending it or processing these.
i.e. ccd calls us with it

timja-kainos: I've made the callback controller more generic so it is needed now

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/CallbackController.java
```diff
@@ -0,0 +1,57 @@
...
+    @PostMapping(path = "/{callback-type}")
+    @ApiOperation("Request more time for response arrived on paper")
+    public CallbackResponse callback(
+        @PathVariable("callback-type") String callbackType,
+        @NotNull @RequestBody CallbackRequest callback
+    ) {
+
+        if (callback.getEventId().equals(CaseEvent.MORE_TIME_REQUESTED_PAPER.getValue())) {
+            switch (callbackType) {
+                case "about-to-start":
```
yarpo: I wonder if this shouldn't be enum?

timja-kainos: doesn't seem useful for the code as its only used in one place, but using it a bit in tests so thinking about it

timja-kainos: now using constants

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/exceptions/BadRequestException.java
```diff
@@ -6,4 +6,8 @@ public BadRequestException(String message, Exception ex) {
         super(message, ex);
     }
 
+    public BadRequestException(String message) {
```
rlewan: Is this mapped to 400 HTTP response anywhere? Isn't there anything available from Spring that we could perhaps reuse?

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/MoreTimeRequestedCallbackTest.java
```diff
@@ -0,0 +1,160 @@
...
+        MvcResult mvcResult = makeRequest(ABOUT_TO_START_CALLBACK, LocalDate.now().plusDays(3), true)
+            .andExpect(status().isOk())
+            .andReturn();
+
+        AboutToStartOrSubmitCallbackResponse response = deserializeObjectFrom(
+            mvcResult,
+            AboutToStartOrSubmitCallbackResponse.class
+        );
+
+        assertThat(response.getErrors()).hasSize(1);
```
rlewan: Maybe it would be worth to assert the error's content here?

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -176,6 +188,58 @@ public Claim requestMoreTimeForResponse(String externalId, String authorisation)
         return claim;
     }
 
+    @SuppressWarnings("unchecked")
+    public AboutToStartOrSubmitCallbackResponse requestMoreTimeOnPaper(
```
rlewan: `ClaimService` is getting quite big, maybe it would be worth moving this to another class? Also this is CCD related.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/CallbackController.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @PostMapping(path = "/{callback-type}")
+    @ApiOperation("Handles all callbacks from CCD")
+    public CallbackResponse callback(
+        @PathVariable("callback-type") String callbackType,
+        @NotNull @RequestBody CallbackRequest callback
+    ) {
+
+        if (callback.getEventId().equals(CaseEvent.MORE_TIME_REQUESTED_PAPER.getValue())) {
```
rlewan: Maybe it would be cleaner to have a separate controller method per event? I.e. not use a path variable.

src/main/java/uk/gov/hmcts/cmc/claimstore/utils/CCDCaseDataToClaim.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public Claim to(Long caseId, Map<String, Object> data) {
+        Map<String, Object> tempData = new HashMap<>(data);
+        tempData.put("id", caseId);
+
+        CCDCase ccdCase = convertToCCDCase(tempData);
+        return caseMapper.from(ccdCase);
+    }
+
+    private CCDCase convertToCCDCase(Map<String, Object> mapData) {
```
rlewan: I think Jackson allows you to specify types you convert between, so you won't need to do double mapping. I.e. map the `Map` to an object.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/CallbackController.java
timja-kainos: I don't think so, this is configured in a spreadsheet and the typo risk is higher than the added complexity IMO

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
timja-kainos: 270 lines, not sure if its worth it.

pragnesh-patel:
Checks haven’t all passed yet

PR #367, opened 2018-05-22
ROC-3664: Notify claimant that defendant is keen to use free mediation service

dharmendrak:
Jira link: https://tools.hmcts.net/jira/browse/ROC-3664


src/main/java/uk/gov/hmcts/cmc/claimstore/services/notifications/DefendantResponseNotificationService.java
```diff
@@ -84,7 +85,22 @@ public void notifyClaimant(
...
-        notify(claim.getSubmitterEmail(), getEmailTemplates().getClaimantResponseIssued(), parameters, reference);
+
+        String emailTemplate = getClaimantEmailTemplate(claim);
+
+        notify(claim.getSubmitterEmail(), emailTemplate, parameters, reference);
+    }
+
+    private String getClaimantEmailTemplate(Claim claim) {
+        if (claim.getResponse().isPresent()) {
+            Response response = claim.getResponse().get();
```
timja-kainos: please don't use `.get()` on optional if it can be avoided, use .orElseThrow here as you're in a response class it is guaranteed you'll have one

src/main/java/uk/gov/hmcts/cmc/claimstore/services/notifications/DefendantResponseNotificationService.java
```diff
@@ -84,7 +85,22 @@ public void notifyClaimant(
...
+
+            if (response.getFreeMediation().isPresent() && response.getFreeMediation().get().equals(YesNoOption.YES)) {
```
timja-kainos: I think `response.getMediation().orElse(NO)` in a variable before this would be cleaner, then you just check if its yes

src/main/java/uk/gov/hmcts/cmc/claimstore/config/properties/notifications/EmailTemplates.java
```diff
@@ -63,6 +63,9 @@
     @NotEmpty
     private String offerCounterSignedEmailToOtherParty;
 
+    @NotEmpty
+    private String freeMediationAcceptedEmailToClaimant;
```
geek-soft: Would it be possible to have one template with optional content? Notify documentation says:

```
Optional content

Use double brackets and ‘??’ to define optional content.

For example if you only want to show something to people who are under 18:

((under18??Please get your application signed by a parent or guardian.))
For each person you send this message to, specify ‘yes’ or ‘no’ to show or hide this content.
```

geek-soft: If not possible lets just rename property to be `claimantResponseWithMediationIssued` or to follow other pattern `responseWithMediationIssuedEmailToClaimant`. My point here is that the pain event is still response submission, `freeMediationAccepted` is a bit confusion for me.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/notifications/DefendantResponseNotificationService.java
```diff
@@ -143,11 +159,14 @@ public void logNotificationFailure(
         parameters.put(DEFENDANT_NAME, claim.getClaimData().getDefendant().getName());
         parameters.put(FRONTEND_BASE_URL, notificationsProperties.getFrontendBaseUrl());
         parameters.put(CLAIM_REFERENCE_NUMBER, claim.getReferenceNumber());
+
         parameters.put(MEDIATION_DECISION_DEADLINE, formatDate(decisionDeadline));
+        parameters.put(MEDIATION_RESPONSE_DEADLINE, Formatting.formatDate(decisionDeadline));
```
geek-soft: It looks like this is duplicate. A `MEDIATION_DECISION_DEADLINE` has the same value and sounds same.

src/main/java/uk/gov/hmcts/cmc/claimstore/config/properties/notifications/EmailTemplates.java
```diff
@@ -63,6 +63,9 @@
     @NotEmpty
     private String offerCounterSignedEmailToOtherParty;
 
+    @NotEmpty
+    private String claimantResponseWithMediationIssued;
```
geek-soft: Can we group notification together so template without mediation and with mediation are kept together.

Also applying the same naming convention will help.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/notifications/DefendantResponseNotificationService.java
```diff
@@ -82,9 +81,21 @@ public void notifyClaimant(
...
+        Map<String, String> parameters = aggregateParams(claim, response);
+
+        String emailTemplate = getClaimantEmailTemplate(response);
+
+        notify(claim.getSubmitterEmail(), emailTemplate, parameters, reference);
+    }
+
+    private String getClaimantEmailTemplate(Response response) {
+        YesNoOption mediation = response.getFreeMediation().orElse(YesNoOption.NO);
+        if (mediation.equals(YesNoOption.YES)) {
```
geek-soft: You can do `==`

pragnesh-patel:
@dharmendrak please see the Codacy failure above: 
src/aat/java/uk/gov/hmcts/cmc/claimstore/tests/functional/RespondToClaimTest.java
JUnit tests should include assert() or fail()
41
public void shouldBeAbleToSuccessfullySubmitFreeMediationRequestOnDefence() {

pragnesh-patel:
Also checking locally the new text (copy) in the mediation email did not show - checked with Abi

PR #374, opened 2018-05-31
ROC-3023 Attach sealed claim to case

timja-kainos:
https://tools.hmcts.net/jira/browse/ROC-3023

### Change description ###
Fix up the integration and start getting it to work on CCD

```
[ ] Yes
[ ] No
```


timja-kainos:
@dharmendrak if it should be named like that then the class should be called that, and not having to name vars differently all over the place, the variable name should only need to be changed if it adds value

PR #388, opened 2018-06-18
ROC-3800 Create DB config for scheduler DB

timja-kainos:


infrastructure/main.tf
```diff
@@ -71,6 +71,19 @@ data "azurerm_key_vault_secret" "oauth_client_secret" {
   vault_uri = "${data.azurerm_key_vault.cmc_key_vault.vault_uri}"
 }
 
+module "scheduler-database" {
+  source = "git@github.com:hmcts/moj-module-postgres?ref=cnp-449-tactical"
+  product = "${var.product}-scheduler"
```
geek-soft: Can we make `"${var.product}-scheduler-db"` to make it easier to find in Azure Portal by name?

timja-kainos: 3. says no https://tools.hmcts.net/confluence/display/CNP/Naming+Conventions

geek-soft: Alright  :)

PR #391, opened 2018-06-20
ROC: 3849 cross field validation

yarpo:
https://tools.hmcts.net/jira/browse/ROC-3849

### Change description ###
Cross-field validation for FullAdmission

```
[x] Yes
[ ] No
```


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidResidence.java
```diff
@@ -0,0 +1,21 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.TYPE, ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Constraint(validatedBy = {ValidResidenceConstraintValidator.class})
+@Documented
+public @interface ValidResidence {
```
sabahirfan: Prefer to create a class level reusable constraint which can validate based on some conditional property. We can apply same on other similar places. 

Something like:
```
@ConditionalRequired(fieldName = "type", values = {"OTHER"}, required = {"otherDetail"})
public class Residence {
    @NotNull
    private final ResidenceType type;
    private final String otherDetail;
...
}
```

and the marker for validator would be like:

```
@Target({ElementType.TYPE})
@Retention(RetentionPolicy.RUNTIME)
@Constraint(validatedBy = {ConditionalRequiredValidator.class})
public @interface ConditionalRequired {

    String message() default "Field is required.";
    Class<?>[] groups() default {};
    Class<? extends Payload>[] payload() default {};

    String fieldName();
    String[] required();
    String[] values();
}
```

timja-kainos: Then there's no type safety and it'll break at runtime if you refactor, I would prefer against this

sabahirfan: The unit tests for these models will catch if we miss anything during refactoring. Also we don't change the field names very often. IMHO reusable constraint will be more appropriate and will reduce duplicate code.

timja-kainos: they may not catch them though, and creating such a validator leads to the chance that it will be used on some optional field which isn't covered fully by tests,
It's unneeded complexity and risk imo

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidChildConstraintValidator.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private boolean validateOlderChild(Child child, ConstraintValidatorContext context) {
+        boolean valid = true;
+        String type = BETWEEN_16_AND_19.toString();
+
+        Optional<Integer> numberOfChildrenLivingWithYou = child.getNumberOfChildrenLivingWithYou();
+        int valueNoOfChildrenLivingWithYou = numberOfChildrenLivingWithYou.orElse(0);
+
+        if (!numberOfChildrenLivingWithYou.isPresent()) {
```
geek-soft: I think it would be cleaner if we had:

```if (!child.getNumberOfChildrenLivingWithYou())```

And then below we would declare:

```int numberOfChildrenLivingWithYou = child.getNumberOfChildrenLivingWithYou().orElse(0);```

File: domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidEmploymentConstraintValidator.java [hunk 3]
```diff
@@ -0,0 +1,46 @@
...
+                );
+                valid = false;
+            }
+        }
+
+        if (employment.getSelfEmployment().isPresent() || employment.getEmployers().size() > 0) {
+            if (employment.getUnemployment().isPresent()) {
+                setValidationErrors(
+                    context,
+                    "selfEmployment",
```
geek-soft: It should be `unemployment` here.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidFullAdmissionConstraintValidator.java
```diff
@@ -0,0 +1,103 @@
...
+            default:
+                return false;
+        }
+    }
+
+    private boolean validateImmediately(FullAdmissionResponse value, ConstraintValidatorContext context) {
+        boolean valid = true;
+        String immediately = IMMEDIATELY.getDescription();
+
+        if (value.getPaymentDate().isPresent()) {
```
geek-soft: Please remove, I will be sending payment date for immediate payment as it will get calculated as `today + 5 days`

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidFullAdmissionConstraintValidator.java
```diff
@@ -0,0 +1,103 @@
...
+            setValidationErrors(context, Fields.PAYMENT_DATE, mayNotBeNullError("paymentType", bySetDate));
+            valid = false;
+        }
+
+        if (value.getRepaymentPlan().isPresent()) {
+            setValidationErrors(context, Fields.REPAYMENT_PLAN, mayNotBeProvidedError("paymentType", bySetDate));
+            valid = false;
+        }
+
+        if (!value.getStatementOfMeans().isPresent()) {
```
geek-soft: We cannot have it here since SoM is only for individuals and sole traders.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidFullAdmissionConstraintValidator.java
```diff
@@ -0,0 +1,103 @@
...
+            setValidationErrors(context, Fields.PAYMENT_DATE, mayNotBeProvidedError("paymentType", instalments));
+            valid = false;
+        }
+
+        if (!value.getRepaymentPlan().isPresent()) {
+            setValidationErrors(context, Fields.REPAYMENT_PLAN, mayNotBeNullError("paymentType", instalments));
+            valid = false;
+        }
+
+        if (!value.getStatementOfMeans().isPresent()) {
```
geek-soft: We cannot have it here since SoM is only for individuals and sole traders.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidUnemploymentConstraintValidator.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    public boolean isValid(Unemployment unemployment, ConstraintValidatorContext context) {
+        boolean valid = true;
+        boolean isRetiredPopulated = unemployment.isRetired();
+        boolean otherPopulated = unemployment.getOther().isPresent();
+        boolean unemployedPopulated = unemployment.getUnemployed().isPresent();
+
+        if (isRetiredPopulated) {
+            if (otherPopulated) {
+                setValidationErrors(context, Fields.OTHER, mayNotBeProvidedError(Fields.OTHER, Fields.IS_RETIRED));
```
geek-soft: This should be `mayNotBeProvidedError(Fields.OTHER` replaced with `mayNotBeProvidedError('unemployment'`.

We want to say here that other is not allowed when unemployment is set to retired.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidUnemploymentConstraintValidator.java
```diff
@@ -0,0 +1,68 @@
...
+                valid = false;
+            }
+
+            if (unemployedPopulated) {
+                setValidationErrors(
+                    context, Fields.UNEMPLOYED, mayNotBeProvidedError(Fields.UNEMPLOYED, Fields.IS_RETIRED)
```
geek-soft: The same here.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidUnemploymentConstraintValidator.java
```diff
@@ -0,0 +1,68 @@
...
+                );
+                valid = false;
+            }
+        }
+
+        if (otherPopulated) {
+            if (isRetiredPopulated) {
+                setValidationErrors(context, Fields.IS_RETIRED, mayNotBeProvidedError(Fields.IS_RETIRED, Fields.OTHER));
```
geek-soft: The same here.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidUnemploymentConstraintValidator.java
```diff
@@ -0,0 +1,68 @@
...
+                valid = false;
+            }
+
+            if (unemployedPopulated) {
+                setValidationErrors(context, Fields.UNEMPLOYED, mayNotBeProvidedError(Fields.UNEMPLOYED, Fields.OTHER));
```
geek-soft: The same here.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidUnemploymentConstraintValidator.java
```diff
@@ -0,0 +1,68 @@
...
+                valid = false;
+            }
+        }
+
+        if (unemployedPopulated) {
+            if (isRetiredPopulated) {
+                setValidationErrors(
+                    context, Fields.IS_RETIRED, mayNotBeProvidedError(Fields.IS_RETIRED, Fields.UNEMPLOYED)
```
geek-soft: The same here.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidUnemploymentConstraintValidator.java
```diff
@@ -0,0 +1,68 @@
...
+                );
+                valid = false;
+            }
+
+            if (otherPopulated) {
+                setValidationErrors(context, Fields.OTHER, mayNotBeProvidedError(Fields.OTHER, Fields.UNEMPLOYED));
```
geek-soft: The same here.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidFullAdmissionConstraintValidator.java
yarpo: ok

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidChildConstraintValidator.java
yarpo: ok

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidEmploymentConstraintValidator.java
yarpo: ok

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidFullAdmissionConstraintValidator.java
yarpo: I have changed the logic. Now, `paymentDate` is mandatory.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidFullAdmissionConstraintValidator.java
yarpo: removed

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidFullAdmissionConstraintValidator.java
yarpo: removed

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidUnemploymentConstraintValidator.java
yarpo: done

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidUnemploymentConstraintValidator.java
yarpo: done

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidUnemploymentConstraintValidator.java
yarpo: done

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidUnemploymentConstraintValidator.java
yarpo: done

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/utils/ConstraintsUtils.java
```diff
@@ -0,0 +1,29 @@
+package uk.gov.hmcts.cmc.domain.constraints.utils;
+
+import javax.validation.ConstraintValidatorContext;
+
+public class ConstraintsUtils {
```
timja-kainos: nice

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/statementofmeans/Child.java
```diff
@@ -22,8 +23,10 @@
     private final AgeGroupType ageGroupType;
 
     @NotNull
+    @Min(0)
```
timja-kainos: i prefer if the smaller annotations go on top, i find it easier to read but up to you 👍

yarpo: ok

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidResidence.java
sabahirfan: Well, some one can still misuse any of the hibernate validators on optional!

I still think `ValidResidence`, `ValidExpense` and `ValidIncome` producing duplication and can be done by one reusable validator. @geek-soft what are your thoughts on this?

PR #394, opened 2018-06-20
Remove ssl from localhost test properties

dharmendrak:


src/aat/resources/environment.properties
```diff
@@ -1,4 +1,4 @@
-spring.datasource.scheduler.url=jdbc:postgresql://localhost:5431/scheduler?ssl
+spring.datasource.scheduler.url=jdbc:postgresql://localhost:5431/scheduler
```
geek-soft: Does it really matter?

timja-kainos: no it doesn't

geek-soft: I guess no point merging this? Shame we need to add these props for AAT test runner.

PR #415, opened 2018-07-16
ROC-4098 Renaming payment option

dharmendrak:


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CCDPaymentOption.java
```diff
@@ -2,7 +2,7 @@
 
 public enum CCDPaymentOption {
     IMMEDIATELY("immediately"),
-    FULL_BY_SPECIFIED_DATE("on or before specified date"),
+    BY_SPECIFIED_DATE("on or before specified date"),
```
yarpo: I suppose you need to update this as well:

src/integration-test/tests/citizen/ccj/pages/payment-options.ts:5

```
const fields = {
  paymentOption: {
    Immediate: 'input[id=optionIMMEDIATELY]',
    Instalments: 'input[id=optionINSTALMENTS]',
    FullBySetDate: 'input[id=optionFULL_BY_SPECIFIED_DATE]'
  }
}
```

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/PaymentOption.java
```diff
@@ -2,6 +2,7 @@
 
 public enum PaymentOption {
     IMMEDIATELY("Immediately"),
+    BY_SPECIFIED_DATE("By a set date"),
     FULL_BY_SPECIFIED_DATE("By a set date"),
```
sabahirfan: I suppose this one we are keeping for backward compatibility only? 

Also on `citizen-frontend` we just have `FULL_BY_SPECIFIED_DATE` how that would work with new `BY_SPECIFIED_DATE` during JSON mapping in Request/Response?

ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CCDPaymentOption.java
dharmendrak: there is another PR for it mate

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/PaymentOption.java
dharmendrak: get on paymentOption will always return BY_SET_DATE

ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CCDPaymentOption.java
yarpo: why? It's the same git repo

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/PaymentOption.java
sabahirfan: yes, that's what I meant and front-end will fail to map this new enum value

dharmendrak: there is a PR for frontend

sabahirfan: Great.

sabahirfan:
Front-end PR: https://github.com/hmcts/cmc-citizen-frontend/pull/676

PR #417, opened 2018-07-19
Rename field in expense model

timja-kainos:


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/constraints/ValidExpenseConstraintValidator.java
```diff
@@ -12,7 +12,7 @@
 public class ValidExpenseConstraintValidator implements ConstraintValidator<ValidExpense, Expense> {
 
     public static class Fields {
-        public static final String OTHER_EXPENSE = "otherExpense";
+        public static final String OTHER_EXPENSE = "otherName";
```
sabahirfan: `OTHER_EXPENSE_NAME`  rather than `OTHER_EXPENSE` ?

timja-kainos:
@sabahirfan not touching ccd other than renaming the model, we'll deal with it when we turn it back on

PR #444, opened 2018-08-15
ROC-4207 Claimant response

timja:
https://tools.hmcts.net/jira/browse/ROC-4207

Save claimant response - future PRs for validation and emails

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/CaseMapper.java
+++ b/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/CaseMapper.java
@@ -134,6 +134,8 @@ public class CaseMapper implements Mapper<CCDCase, Claim> {
             settlement,
             fromNullableUTCtoLocalZone(ccdCase.getSettlementReachedAt()),
             mapSealedClaimDocument(ccdCase.getSealedClaimDocument()),
+            null,
+            null,
             null
         );
     }
--- a/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
+++ b/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
@@ -211,8 +211,8 @@ public class Claim {
         return Optional.ofNullable(claimantResponse);
     }
 
-    public LocalDateTime getClaimantRespondedAt() {
-        return claimantRespondedAt;
+    public Optional<LocalDateTime> getClaimantRespondedAt() {
+        return Optional.ofNullable(claimantRespondedAt);
     }
 
     public List<String> getFeatures() {
--- a/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
+++ b/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
@@ -7,6 +7,7 @@ import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
 import uk.gov.hmcts.cmc.domain.amount.TotalAmountCalculator;
+import uk.gov.hmcts.cmc.domain.models.claimantresponse.ClaimantResponse;
 import uk.gov.hmcts.cmc.domain.models.offers.Settlement;
 import uk.gov.hmcts.cmc.domain.models.response.Response;
 
@@ -51,6 +52,8 @@ public class Claim {
     private final LocalDateTime settlementReachedAt;
     private final URI sealedClaimDocument;
     private final List<String> features;
+    private final LocalDateTime claimantRespondedAt;
+    private final ClaimantResponse claimantResponse;
 
     @SuppressWarnings("squid:S00107") // Not sure there's a lot fo be done about removing parameters here
     @JsonCreator
@@ -75,7 +78,9 @@ public class Claim {
         Settlement settlement,
         LocalDateTime settlementReachedAt,
         URI sealedClaimDocument,
-        List<String> features
+        List<String> features,
+        LocalDateTime claimantRespondedAt,
+        ClaimantResponse claimantResponse
     ) {
         this.id = id;
         this.submitterId = submitterId;
@@ -98,6 +103,8 @@ public class Claim {
         this.settlementReachedAt = settlementReachedAt;
         this.sealedClaimDocument = sealedClaimDocument;
         this.features = features;
+        this.claimantRespondedAt = claimantRespondedAt;
+        this.claimantResponse = claimantResponse;
     }
 
     public Long getId() {
@@ -198,6 +205,14 @@ public class Claim {
 
     public Optional<BigDecimal> getTotalInterest() {
         return TotalAmountCalculator.calculateInterestForClaim(this);
+    }
+
+    public Optional<ClaimantResponse> getClaimantResponse() {
+        return Optional.ofNullable(claimantResponse);
+    }
+
+    public LocalDateTime getClaimantRespondedAt() {
+        return claimantRespondedAt;
     }
 
     public List<String> getFeatures() {
--- a/domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleClaimantResponse.java
+++ b/domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleClaimantResponse.java
@@ -1,0 +1,76 @@
+package uk.gov.hmcts.cmc.domain.models.sampledata;
+
+import uk.gov.hmcts.cmc.domain.models.claimantresponse.ClaimantResponse;
+import uk.gov.hmcts.cmc.domain.models.claimantresponse.ResponseAcceptation;
+import uk.gov.hmcts.cmc.domain.models.claimantresponse.ResponseRejection;
+
+import java.math.BigDecimal;
+
+public abstract class SampleClaimantResponse<T extends SampleClaimantResponse<T>> {
+
+    public static ClaimantResponse validDefaultAcceptation() {
+        return ClaimantResponseAcceptation.builder().build();
+    }
+
+    public static ClaimantResponse validDefaultRejection() {
+        return ClaimantResponseRejection.builder().build();
+    }
+
+    public static class ClaimantResponseAcceptation extends SampleClaimantResponse<ClaimantResponseAcceptation> {
+
+        private BigDecimal amountPaid = BigDecimal.TEN;
+
+        public static ClaimantResponseAcceptation builder() {
+            return new ClaimantResponseAcceptation();
+        }
+
+        public ClaimantResponseAcceptation withAmountPaid(BigDecimal amountPaid) {
+            this.amountPaid = amountPaid;
+            return this;
+        }
+
+        @Override
+        public ClaimantResponse build() {
+            return ResponseAcceptation.builder()
+                .amountPaid(amountPaid)
+                .build();
+        }
+    }
+
+    public static class ClaimantResponseRejection extends SampleClaimantResponse<ClaimantResponseAcceptation> {
+
+        private BigDecimal amountPaid = BigDecimal.TEN;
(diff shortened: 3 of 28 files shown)


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/CaseMapper.java
```diff
@@ -134,6 +134,8 @@ public Claim from(CCDCase ccdCase) {
             settlement,
             fromNullableUTCtoLocalZone(ccdCase.getSettlementReachedAt()),
             mapSealedClaimDocument(ccdCase.getSealedClaimDocument()),
+            null,
```
sabahirfan: I will prefer to replace this call with Builder we have provided by lombok. Calling a constructor with very long arguments is not clean

timja: The thing I like about a constructor here is that we know all the arguments are called, we want to make sure that we map every field to CCD or not as is our current situation

sabahirfan: We are not mapping all here as passing null for the time being. Even if we have to map all, does not mean using builder is bad here. 

We are having more than 22 constructor arguments to initialise claim object, IMHO using constructor is more error prone as one can pass arguments in wrong order and have totally incorrect data. calling builder will make sure we are mapping the right fields. It's more readable as well.

timja: But when you add more fields it forces you to handle it, making it a compile time error.
Which is a good thing here IMO.

I take your point though, I would be interested in others opinion on this.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
```diff
@@ -200,6 +207,14 @@ public LocalDate getServiceDate() {
         return TotalAmountCalculator.calculateInterestForClaim(this);
     }
 
+    public Optional<ClaimantResponse> getClaimantResponse() {
+        return Optional.ofNullable(claimantResponse);
+    }
+
+    public LocalDateTime getClaimantRespondedAt() {
```
sabahirfan: shouldn't this be optional as well?

domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleClaimantResponse.java
```diff
@@ -0,0 +1,76 @@
+package uk.gov.hmcts.cmc.domain.models.sampledata;
+
+import uk.gov.hmcts.cmc.domain.models.claimantresponse.ClaimantResponse;
+import uk.gov.hmcts.cmc.domain.models.claimantresponse.ResponseAcceptation;
+import uk.gov.hmcts.cmc.domain.models.claimantresponse.ResponseRejection;
+
+import java.math.BigDecimal;
+
+public abstract class SampleClaimantResponse<T extends SampleClaimantResponse<T>> {
```
sabahirfan: If we use the `Builder` provided by `lombok`, we will not require these sample classes with explicit builder pattern. I know we have a lot of these already but we should try to clean up, and not add more explicit builders.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
timja: Pattern seems to be these dates aren't optional

domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleClaimantResponse.java
timja: The samples provide the required fields pre-filled out to make tests easier to write

src/main/java/uk/gov/hmcts/cmc/claimstore/exceptions/ClaimantResponseAlreadySubmittedException.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.cmc.claimstore.exceptions;
+
+public class ClaimantResponseAlreadySubmittedException extends ConflictException {
+    public ClaimantResponseAlreadySubmittedException(long claimId) {
```
sabahirfan: will prefer not to use DB generated id here, can be UUID or similar which will work for CCD as well

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CaseRepository.java
```diff
@@ -24,6 +25,8 @@
 
     void saveDefendantResponse(Claim claim, String defendantEmail, Response response, String authorization);
 
+    void saveClaimantResponse(long claimId, ClaimantResponse response, String authorization);
```
sabahirfan: same here, prefer not to use DB generated id in contract here as won't work for CCD. May be externalId (UUID) is alternative

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/mapping/ClaimMapper.java
```diff
@@ -82,4 +84,12 @@ private Settlement toNullableSettlement(String input) {
         }
     }
 
+    private <T> T toNullableEntity(String input, TypeReference<T> entityClass) {
```
sabahirfan: I can see this method duplicated in two places, is there an option to reuse after moving in common place?

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/SaveClaimantResponseTest.java
```diff
@@ -0,0 +1,109 @@
...
+    properties = {
+        "core_case_data.api.url=false"
+    }
+)
+public class SaveClaimantResponseTest extends BaseIntegrationTest {
+
+    private Claim claim;
+
+    @Before
+    public void setup() {
```
sabahirfan: oops, should be `setUp()`

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/mapping/ClaimMapper.java
timja: Its duplicated in the migration project which currently has everything duplicated, out of scope for this PR

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/SaveClaimantResponseTest.java
timja: doesn't really matter does it?

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
timja: no, i mean look at all the others

src/main/java/uk/gov/hmcts/cmc/claimstore/exceptions/ClaimantResponseAlreadySubmittedException.java
timja: I can change it, but
1. Its consistent with defendant response
2. it would work just fine on ccd

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CaseRepository.java
timja: why wouldn't it work on ccd?

domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleClaimantResponse.java
timja: I tried but the lombok builder doesn't work with the sample class builder method.
The builder constructs a sample object and not the actual object.

The sample project doesn't even have lombok as a dependency right now, and the sample models do use the builders on the main domain models.

The approach I have done is consistent with the full and partial admission changes which were done very recently

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CaseRepository.java
timja: I've checked, and id is correct here.
On an update to CCD we need the ID so that we don't make any additional calls to retrieve it,
the settlement code does the same as this although it passed the whole claim then gets the ID straight out and passes it along to the service.

The whole object is not needed here

sabahirfan: We don't need to pass whole object, just replace DB generated id with externalId, would be like:
```
 void saveClaimantResponse(String externalId, ClaimantResponse response, String authorization);
```

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/mapping/ClaimMapper.java
sabahirfan: agree, we can leave it as it is

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/SaveClaimantResponseTest.java
sabahirfan: just camel case naming convention, nothing more than that

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
sabahirfan: we might need to clean others as well at some point. prefer to use `Optional` instead of returning `null`

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CaseRepository.java
timja: no no no, then you have to call CCD again to retrieve the claim object you had just before.
We need the ID for the next ccd call. I've passed the required information

domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleClaimantResponse.java
sabahirfan: ok, leave it as it is, I will clean this with others at some point. 

The point was, we don't need to introduce Builder in sample classes, instead just use the Builders from domain classes to generate sample data.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CaseRepository.java
sabahirfan: There will be no additional call to CCD or any repository. We are also passing `externalId` to the API endpoint. 
```
@Transactional(transactionManager = "transactionManager")
    public void save(
        String externalId,
        String claimantId,
        ClaimantResponse response,
        String authorization
    ) {
        Claim claim = claimService.getClaimByExternalId(externalId, authorization);
        claimantResponseRule.assertCanBeRequested(claim, claimantId);

        caseRepository.saveClaimantResponse(claim.getId(), response, authorization); 
        // The above will change to as below, no additional call to retrieve claim
       caseRepository.saveClaimantResponse(claim.getExternalId(), response, authorization); 

        appInsights.trackEvent(getAppInsightsEvent(response), claim.getReferenceNumber());
    }
```

timja: discussed on slack, TLDR id is needed for calling to CCD. The contract here is correct

PR #447, opened 2018-08-16
ROC-4257 Correct interest end date for RPA

ellie-harrison:
https://tools.hmcts.net/jira/browse/ROC-4257

### Change description ###
Updated the ClaimDataContentProvider used for PDF generation to use the correct interest end date so the total claim amount matches the RPA JSON generated at the same time for staff notification.

```
[ ] Yes
[x] No
```


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/amount/TotalAmountCalculator.java
```diff
@@ -32,6 +32,10 @@ private TotalAmountCalculator() {
         return Optional.ofNullable(calculateTotalAmount(claim, LocalDate.now(), false));
     }
 
+    public static Optional<BigDecimal> amountWithInterestUntilSubmitDate(Claim claim) {
```
timja: submit date and issued on are 2 different dates, which one do you want?

PR #448, opened 2018-08-17
ROC-4207 Save claimant response send email

timja:
https://tools.hmcts.net/jira/browse/ROC-4207

### Change description ###
Send claimant response saved email

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/events/claimantresponse/ClaimantResponseActionsHandler.java
```diff
@@ -0,0 +1,49 @@
...
+        NotificationsProperties notificationsProperties
+    ) {
+        this.notificationService = notificationService;
+        this.notificationsProperties = notificationsProperties;
+    }
+
+    @EventListener
+    public void sendNotificationToDefendant(ClaimantResponseEvent event) {
+        Claim claim = event.getClaim();
+        this.notificationService.sendMail(
```
sabahirfan: I think we can reuse `NotificationToDefendantService` to send this email, as code below is exactly sending the same email template.

dharmendrak: I agree with you Sabah, however if you have to change the template, then it need to reworked as now.  if Tim is changing for this, then he need to refactor all places even in ccd.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/notifications/NotificationReferenceBuilder.java
```diff
@@ -165,4 +165,17 @@ public static String referenceForDefendant(String claimReferenceNumber, String o
             return reference(TEMPLATE, DEFENDANT, otherParty.toLowerCase(), claimReferenceNumber);
         }
     }
+
+    public static class ClaimantResponse {
```
sabahirfan: We don't need this new class as we already have similar static inner class in same file called `ClaimantResponseSubmitted`.

timja: its gone now

src/main/java/uk/gov/hmcts/cmc/claimstore/events/claimantresponse/ClaimantResponseActionsHandler.java
timja: done, didn't see that before

sabahirfan: Thanks @timja 
 @dharmendrak  `CCJ` cleanup can be done in separate PR IMO

PR #449, opened 2018-08-17
ROC-4262 Use docker in the pipeline

timja:
https://tools.hmcts.net/jira/browse/ROC-4262

Enable docker in the pipeline and AKS

```
[ ] Yes
[ ] No
```


Dockerfile
```diff
@@ -1,11 +1,11 @@
 FROM openjdk:8-jre-alpine
```
timwebster9: please use the cnp-java-base image: https://github.com/hmcts/cnp-java-base

src/kubernetes/.lint.yml
```diff
@@ -0,0 +1,2 @@
+rules:
+  line-length: disable
```
timwebster9: how is this used?  Could we add this to the CI somehow?

src/kubernetes/deployment.template.yaml
```diff
@@ -0,0 +1,161 @@
...
+              value: "${STAFF_EMAIL}"
+            - name: CLAIM_STORE_TEST_SUPPORT_ENABLED
+              value: 'true'
+            - name: FEATURE_TOGGLES_EMAILTOSTAFF
+              value: 'true'
+            - name: ROOT_APPENDER
+              value: 'CMC'
+            - name: APPINSIGHTS_INSTRUMENTATIONKEY
+              value: 'TODO'
+
```
timwebster9: You could add a configmap to this file and reference all variables from that, in preparation for https://tools.hmcts.net/jira/browse/CNP-758 if we go down that route

src/kubernetes/.lint.yml
timja: currently locally with:
`yamllint -c src/kubernetes/.lint.yml src/kubernetes/deployment.template.yaml`

src/kubernetes/deployment.template.yaml
timja: happy to do it when the library supports it

src/kubernetes/.lint.yml
timwebster9: ok so it's an external tool? Maybe we could add it to the docker image that kubectl/az cli run in, and run it before the `kubectl apply`.  Only if it adds value though - it's a bit late to run it at this point.

Don't really want to get into having to add stuff to the agent VM and all that :-)

timja: should just be able to run it in an image, it should run in the test section i think well before this bit if the flag is enabled. but we can probably leave that for this PR

src/kubernetes/deployment.template.yaml
timwebster9: if it's part of this same file it'll just work.  See here for example: https://github.com/hmcts/jenkins-container-config/blob/65aac84384bd3e4b45f1027349b682b3f3def387/aks/jenkins.yaml#L31

timwebster9: can move it to another file once the library is updated to use different files per env

PR #451, opened 2018-08-20
[PREVIEW] ROC-4278: Update legal terms in generated PDF docs

sabahirfan:
[ROC-4278](https://tools.hmcts.net/jira/browse/ROC-4278)

### Change description ###
This PR provides dynamic form numbers for defendant response and settlement agreement, along with few static content changes for legal terms.

- [ ] Yes
- [x] No


src/main/resources/staff/templates/document/defendantPinLetter.html
```diff
@@ -51,7 +51,7 @@
         </td>
         <td class="no-spacing">
           <p class="address" style="float:right">
-            County Court Business Centre Online Court<br/>
+            County Court Business Centre Online Civil Money Claims<br/>
```
geek-soft: Sometimes we have these in two lines, sometimes in one line. Is there any rule? I guess we should aim to make it consistent and have two lines.

sabahirfan: After demo to PO, decided to break in two line where we have image logo, otherwise it should be one line as per PO (Marc)

sabahirfan: Discussed with PO and he agrees to make these consistent on two lines.

PR #459, opened 2018-08-29
Update idam url

robertparkinson:


infrastructure/prod.tfvars
```diff
@@ -1,4 +1,4 @@
-idam_api_url = "https://prod-idamapi.reform.hmcts.net"
+idam_api_url = "http://idam-api.platform.hmcts.net/"
```
geek-soft: Remove trailing slash.

robertparkinson: done and added back in the https

timja:
I've confirmed the URL works now

PR #465, opened 2018-09-04
Enable auto merge to demo

sabahirfan:


pragnesh-patel:
So we need Demo to reflect Master branch

PR #466, opened 2018-09-05
Add new sslmode property

timja:
Required after dependency upgrade (breaking change in patch version -.-)

https://github.com/pgjdbc/pgjdbc/blob/master/CHANGELOG.md


ckalista:
Looks good to me 👍

PR #487, opened 2018-09-19
[PREVIEW] Enable ccd  on preview

dharmendrak:


robertparkinson:
we don't use preview any more.

PR #490, opened 2018-09-21
ROC-4346: Add decision type for determination

sabahirfan:
### Change description ###
This PR adds decision type for court determination results.

- [ ] Yes
- [x] No


ckalista:
Hi @sabahirfan,

After further discussions it become more apparent that another slight change is likely to be needed. It should be easy to implement but first I would like to discuss with you the rational on Monday. For now just a tease, I don't think we will need a type but instead we should store the Court Decision as a whole. The flag idea we had won't work for all scenarios since in some cases when the claimant counter offer is accepted we still need to convert the frequency of the instalment plan in terms of the defendants frequency. I will explain this!

Also we should think of persisting the disposable income amount.

* We should persist:
	* PaymentIntention
		* PaymentOption
		* PaymentDate
		* RepaymentPlan
			* InstalmentAmount
			* FirstPaymentDate
			* PaymentSchedule
			* **CompletionDate**
	* CourtDetermination
		* **CourtDecision: PaymentIntention**
		* RejectedReason
		* **DisposableIncome**

src/test/java/uk/gov/hmcts/cmc/claimstore/services/FormaliseResponseAcceptanceServiceTest.java
```diff
@@ -463,7 +496,7 @@ public void formaliseDoesNothingWhenReferredToJudge() {
     }
 
     @Test
-    public void formaliseDoesNothingWhenResponseIsNotAcceptation() {
+    public void doNotFormaliseWhenResponseIsNotAcceptation() {
```
geek-soft: Not part of this story but if formalise is not possible upon rejection then no point accepting both subtypes of `ClaimantResponse` as an argument.

PR #499, opened 2018-09-26
ROC-4413:  Make amountPaid optional and rename DeterminationDecisionType 

sabahirfan:

[ROC-4413](https://tools.hmcts.net/jira/browse/ROC-4413)

### Change description ###
This PR makes `amountPaid` optional and renames  `DeterminationDecisionType` to `DecisionType`

- [ ] Yes
- [x] No

### DIFF
--- a/domain-model/src/test/java/uk/gov/hmcts/cmc/domain/models/claimantresponse/ResponseAcceptationTest.java
+++ b/domain-model/src/test/java/uk/gov/hmcts/cmc/domain/models/claimantresponse/ResponseAcceptationTest.java
@@ -3,6 +3,7 @@ package uk.gov.hmcts.cmc.domain.models.claimantresponse;
 import org.junit.Test;
 import uk.gov.hmcts.cmc.domain.models.sampledata.SampleClaimantResponse;
 
+import java.math.BigDecimal;
 import java.util.Set;
 
 import static java.math.BigDecimal.TEN;
@@ -35,6 +36,19 @@ public class ResponseAcceptationTest {
         Set<String> response = validate(claimantResponse);
 
         assertThat(response).hasSize(0);
+    }
+
+    @Test
+    public void shouldBeInvalidWhenAmountIsNegative() {
+        ClaimantResponse claimantResponse = ResponseAcceptation.builder()
+            .amountPaid(BigDecimal.valueOf(-10))
+            .formaliseOption(REFER_TO_JUDGE)
+            .decisionType(DEFENDANT)
+            .build();
+
+        Set<String> response = validate(claimantResponse);
+
+        assertThat(response).hasSize(1);
     }
 
     @Test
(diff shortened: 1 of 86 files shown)


domain-model/src/test/java/uk/gov/hmcts/cmc/domain/models/claimantresponse/ResponseAcceptationTest.java
```diff
@@ -25,38 +24,12 @@ public void shouldBeSuccessfulValidationForValidResponse() {
...
-            .determinationDecisionType(DEFENDANT)
-            .build();
-
-        Set<String> response = validate(claimantResponse);
-
-        assertThat(response).hasSize(1);
-    }
-
-    @Test
-    public void shouldBeInvalidWhenAmountIsNegative() {
```
geek-soft: I bet that should stay. Negative amounts should be forbidden.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/FormaliseResponseAcceptanceService.java
```diff
@@ -179,7 +179,7 @@ private PaymentIntention acceptedPaymentIntention(ResponseAcceptation responseAc
                     .getCourtDetermination().orElseThrow(IllegalStateException::new)
                     .getCourtDecision();
             default:
-                throw new IllegalStateException("Invalid determination decision type " + determinationDecisionType);
+                throw new IllegalStateException("Invalid determination decision type " + decisionType);
```
geek-soft: Shouldn't comment get updated as well in case class rename?

sabahirfan: agree

domain-model/src/test/java/uk/gov/hmcts/cmc/domain/models/claimantresponse/ResponseAcceptationTest.java
sabahirfan: agree

domain-model/src/test/java/uk/gov/hmcts/cmc/domain/models/claimantresponse/ResponseAcceptationTest.java
```diff
@@ -26,44 +26,31 @@ public void shouldBeSuccessfulValidationForValidResponse() {
     }
 
     @Test
-    public void shouldBeInvalidWhenAmountNotPresent() {
+    public void shouldBeValidWhenAmountIsZero() {
```
geek-soft: You could have kept an old case changing invalid to valid. This way case where amount is null would have been tested as well.

sabahirfan: do we really want to test validation on optional (not annotated) ?

PR #509, opened 2018-10-01
Generate event once claimant response successfully saved

sabahirfan:

[ROC-4413](https://tools.hmcts.net/jira/browse/ROC-4413)

### Change description ###
This PR fixes the issue where event was generated before preserving the data in repository.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
@@ -49,12 +49,12 @@ public class ClaimantResponseService {
         claimantResponseRule.assertCanBeRequested(claim, claimantId);
 
         caseRepository.saveClaimantResponse(claim, response, authorization);
-        eventProducer.createClaimantResponseEvent(claim);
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
```diff
@@ -49,12 +49,12 @@ public void save(
         claimantResponseRule.assertCanBeRequested(claim, claimantId);
 
         caseRepository.saveClaimantResponse(claim, response, authorization);
-        eventProducer.createClaimantResponseEvent(claim);
 
         if (response instanceof ResponseAcceptation && ((ResponseAcceptation) response).getFormaliseOption() != null) {
             formaliseResponseAcceptanceService.formalise(claim, (ResponseAcceptation) response, authorization);
         }
 
+        eventProducer.createClaimantResponseEvent(claim);
```
singh-surjeet: Is there a test case where order of invocation using Mockito can be verified? Also does save needs to happen before formalise?

sabahirfan: Yes save will happen before formalise.

PR #521, opened 2018-10-17
Fixing sonar low code coverage issues

singh-surjeet:
### Change description ###
Adding test case to bump up code coverage over 80%. Will add more test if still under par. 

```
[ ] Yes
[X] No
```


singh-surjeet:
@pragnesh-patel No. You can see @markry-kainos approving it. Code is behind toggle and rest are test cases.

PR #538, opened 2018-10-31
ROC-3809 Alow states paid acceptance with no formalisation

ellie-harrison:
https://tools.hmcts.net/jira/browse/ROC-3809

### Change description ###
Update claimant response controller to allow acceptance of a states paid response which doesn't require formalisation

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
```diff
@@ -51,14 +55,32 @@ public void save(
...
+        Response response = claim.getResponse().orElseThrow(IllegalStateException::new);
+
+        if (ClaimantResponseType.ACCEPTATION.equals(claimantResponse.getType())
+            && !isResponseStatesPaid(response)) {
+            formaliseResponseAcceptanceService.formalise(claim, (ResponseAcceptation) claimantResponse, authorization);
+        }
+    }
+
+    private boolean isResponseStatesPaid(Response response) {
+        switch (response.getResponseType()) {
```
karthikkannankainos: Suggestion :- Use switch if there are more no of cases. For just two types, use if else.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
```diff
@@ -51,14 +55,32 @@ public void save(
...
 
         eventProducer.createClaimantResponseEvent(claim);
         appInsights.trackEvent(getAppInsightsEvent(response), claim.getReferenceNumber());
     }
 
+    private void formaliseResponseAcceptance(ClaimantResponse claimantResponse, Claim claim, String authorization) {
+        Response response = claim.getResponse().orElseThrow(IllegalStateException::new);
+
+        if (ClaimantResponseType.ACCEPTATION.equals(claimantResponse.getType())
+            && !isResponseStatesPaid(response)) {
```
sabahirfan: please add unit tests for this scenario

sabahirfan:
Would be nice if we have functional test for this scenario as well

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
```diff
@@ -51,14 +55,33 @@ public void save(
...
+        formaliseResponseAcceptance(response, claim, authorization);
 
         eventProducer.createClaimantResponseEvent(claim);
         appInsights.trackEvent(getAppInsightsEvent(response), claim.getReferenceNumber());
     }
 
+    private void formaliseResponseAcceptance(ClaimantResponse claimantResponse, Claim claim, String authorization) {
+        Response response = claim.getResponse().orElseThrow(IllegalStateException::new);
+
+        if (ClaimantResponseType.ACCEPTATION.equals(claimantResponse.getType())
```
dharmendrak: use == instead of equals

src/test/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseServiceTest.java
```diff
@@ -54,6 +59,9 @@
     @Mock
     private FormaliseResponseAcceptanceService formaliseResponseAcceptanceService;
 
+    @Rule
```
dharmendrak: looks like it's not used in this class?

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
```diff
@@ -52,14 +56,33 @@ public void save(
...
+    private void formaliseResponseAcceptance(ClaimantResponse claimantResponse, Claim claim, String authorization) {
+        Response response = claim.getResponse().orElseThrow(IllegalStateException::new);
+
+        if (ACCEPTATION == claimantResponse.getType()
+            && !isResponseStatesPaid(response)) {
+            formaliseResponseAcceptanceService.formalise(claim, (ResponseAcceptation) claimantResponse, authorization);
+        }
+    }
+
+    private boolean isResponseStatesPaid(Response response) {
```
sabahirfan: I think we can move this method in some helper class and write it's unit tests as well. It will be reusable.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/utils/ResponseUtils.java
```diff
@@ -0,0 +1,29 @@
...
+        if (response == null) {
+            return false;
+        }
+
+        switch (response.getResponseType()) {
+            case FULL_DEFENCE:
+                return ((FullDefenceResponse) response).getDefenceType() == DefenceType.ALREADY_PAID;
+            case PART_ADMISSION:
+                return ((PartAdmissionResponse) response).getPaymentDeclaration().isPresent();
+            case FULL_ADMISSION:
```
karthikkannankainos: Suggestion - case FULL_ADMISION can be removed. so it defaults to false.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
```diff
@@ -52,14 +54,21 @@ public void save(
...
-        if (ACCEPTATION == response.getType()) {
-            formaliseResponseAcceptanceService.formalise(updatedClaim, (ResponseAcceptation) response, authorization);
-        }
+        formaliseResponseAcceptance(response, updatedClaim, authorization);
 
         eventProducer.createClaimantResponseEvent(updatedClaim);
         appInsights.trackEvent(getAppInsightsEvent(response), claim.getReferenceNumber());
     }
 
+    private void formaliseResponseAcceptance(ClaimantResponse claimantResponse, Claim claim, String authorization) {
```
karthikkannankainos: Since the claim passed to this method is updated with claimantresponse. cant we fetch the claimantresponse from the claim itself? why do we need to pass it as an argument. Happy to be corrected.

sabahirfan: Agree. previously the claim object did't have `ClaimantResponse` at this point, hence we were passing these separately. Now we can clean this up as `updatedClaim` contains the `ClaimantResponse` preserved in database.

@ellie-harrison This can be a separate PR to cleanup.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
```diff
@@ -52,14 +54,21 @@ public void save(
...
+        Response response = claim.getResponse().orElseThrow(IllegalStateException::new);
+
+        if (ACCEPTATION == claimantResponse.getType()
+            && !ResponseUtils.isResponseStatesPaid(response)) {
```
singh-surjeet: @ellie-harrison @sabahirfan @dharmendrak - Is this correct that if payment declaration is present in the case of part admit, claimant cannot formalise claimant response (either as CCJ or Settlement)? One of the AAT functional test is failing on my branch where defendant part admit with payment declaration, so I'm curious if this is correct? Happy to change test if this is the case?

PR #560, opened 2018-11-12
[WIP] ROC-4713 mask PII with sensitive aware (Reflection)ToStringBuilder

andrewwa-kainos:
https://tools.hmcts.net/jira/browse/ROC-4713

### Change description ###
Candidate strategy to prevent `model.toString()` outputting sensitive data

- mark sensitive fields with `@Sensitive`
- call `SensitiveAwareToStringBuilder` instead of `ReflectionToStringBuilder`


sabahirfan:
The masking of sensitive information in logs should be part of the reform-logging library (https://github.com/hmcts/java-logging):
- This would be common use case for all other services in reform program and efforts should not be duplicated.
- This can be achieved by either extending `PatternLayout` and use Regex to mask information (phone, email, address etc.) As this information can be entered by the user in our textarea for explanations.

PR #561, opened 2018-11-12
ROC-4713 Mask sensitive data with aspects

andrewwa-kainos:
https://tools.hmcts.net/jira/browse/ROC-4713

### Change description ###
Candidate strategy to prevent `model.toString()` outputting sensitive data

- mark sensitive fields with `@Sensitive`
- use an aspect to intercept `toString()` calls and sanitise result
  - this example replaces the implementation entirely if any `@Sensitive` fields are found
  - could alternatively search the result and mask any occurrences of the values of any `@Sensitive` fields

### DIFF
--- a/domain-model/build.gradle
+++ b/domain-model/build.gradle
@@ -1,3 +1,12 @@
+buildscript {
+  repositories {
+    jcenter()
+    maven { url 'http://repo.spring.io/plugins-release' }
+  }
+  dependencies {
+  }
+}
+
 plugins {
   id 'java'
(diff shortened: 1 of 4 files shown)


sabahirfan:
👍 
IMHO this AspectJ approach is better choice. We can create Pointcut/s on toString() or other method calls where ever needed. So your PR# 561 approach looks more cleaner.

In other approach, `@Sensative` annotation would not be enough, we need to make sure we call custom toString builder as well.

PR #572, opened 2018-11-19
ROC-4747 - bug fix for pdf exception when defendant is a company

rattur:

https://tools.hmcts.net/jira/browse/ROC-4747

### Change description ###

Bug fix for pdf exception when defendant is a company

```
[ ] Yes
[x ] No
```


src/main/resources/citizen/templates/document/defendantResponseReceipt.html
```diff
@@ -457,7 +457,7 @@
   </div>
 </div>
 
-{% if paymentOption == 'By instalments' or paymentOption == 'By a set date' %}
+{% if (paymentOption == 'By instalments' or paymentOption == 'By a set date') and  defendant.partyType == "INDIVIDUAL" %}
```
singh-surjeet: Are sole traders not dealt in same way? Does that needs to be added to if block to separated out to filter out companies/org?
I know we haven't written response PDF aat/functional test but I think we should write one now making sure PDF is generated with correct contents?

sabahirfan: Also how about checking `statementOfMeans is defined` rather than `partyType` ?

src/main/resources/citizen/templates/document/defendantResponseReceipt.html
```diff
@@ -457,7 +457,7 @@
   </div>
 </div>
 
-{% if paymentOption == 'By instalments' or paymentOption == 'By a set date' %}
+{% if (paymentOption == 'By instalments' or paymentOption == 'By a set date') and  statementOfMeans is defined %}
```
sabahirfan: you need to populate `statementOfMeans`, otherwise you have broken the individual flow.

PR #578, opened 2018-11-23
ROC-4773: Store claim in ccd in parallel thread

dharmendrak:
Jira link:
https://tools.hmcts.net/jira/browse/ROC-4773


src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -220,7 +218,7 @@ private void linkToCase(User defendantUser, User anonymousCaseWorker, String let
         CaseDetails caseDetails = this.updateDefendantIdAndEmail(defendantUser, caseId, defendantId, defendantEmail);
 
         Claim claim = ccdCaseDataToClaim.to(caseDetails.getId(), caseDetails.getData());
-        jobSchedulerService.scheduleEmailNotificationsForDefendantResponse(claim);
+//        jobSchedulerService.scheduleEmailNotificationsForDefendantResponse(claim);
```
ckalista: This should be made conditional based on `ccd_aync_enabled` flag. Only link if the flag is TRUE.

src/kubernetes/deployment.template.yaml
```diff
@@ -88,6 +88,10 @@ spec:
               value: 'true'
             - name: FEATURE_TOGGLES_EMAILTOSTAFF
               value: 'true'
+            - name: FEATURE_TOGGLES_CCD_ENABLED
+              value: 'false'
+            - name: FEATURE_TOGGLES_CCD_ASYNC_ENABLED
+              value: 'false'
```
markry-kainos: remove CORE_CASE_DATA_API_URL from this file

ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/EvidenceMapper.java
```diff
@@ -22,7 +22,7 @@ public EvidenceMapper(EvidenceRowMapper evidenceRowMapper) {
 
     @Override
     public CCDEvidence to(Evidence evidence) {
-        if (evidence == null) {
+        if (evidence == null || evidence.getRows() == null || evidence.getRows().isEmpty()) {
```
srineelesh: why not use something like CollectionUtils.isEmpty(evidence.getRows()) which will check both conditions

PR #648, opened 2019-01-08
ROC- 4739: Def comp/org notification

singh-surjeet:

https://tools.hmcts.net/jira/browse/ROC-4739

### Change description ###

Notify comp/org when claimant rejects repayment plan and it gets referred to judge. 

```
[ ] Yes
[X] No
```


domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleResponse.java
```diff
@@ -107,6 +107,19 @@ public PartAdmissionResponse buildWithPaymentOptionInstallments() {
                 .statementOfMeans(SampleStatementOfMeans.builder().build())
                 .build();
         }
+
+        public PartAdmissionResponse buildWithPaymentOptionInstallmentsAndParty(Party party) {
```
markry-kainos: only 1 L in installments :)

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
```diff
@@ -62,23 +63,25 @@ public ClaimantResponseService(
...
+        Claim updatedClaim = caseRepository.saveClaimantResponse(claim, claimantResponse, authorization);
         claimantResponseRule.isValid(updatedClaim);
-        formaliseResponseAcceptance(response, updatedClaim, authorization);
-        if (isRejectPartAdmitNoMediation(response, updatedClaim)) {
+        formaliseResponseAcceptance(claimantResponse, updatedClaim, authorization);
+        if (isRejectPartAdmitNoMediation(claimantResponse, updatedClaim)) {
             updateDirectionsQuestionnaireDeadline(updatedClaim, authorization);
         }
-        if (!isReferredToJudge(response)) {
+        Response response = claim.getResponse().orElseThrow(IllegalArgumentException::new);
```
markry-kainos: should we do this higher up? - e.g. after `claimantResponseRule.assertCanBeRequested(claim, claimantId);`

singh-surjeet: assertCanBeRequested is doing this check inside as a validation check bit differently but we wouldn't be using response until few lines later and checkstyle is likely to complain. Check `VariableDeclarationUsageDistance`

domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleResponse.java
singh-surjeet: Extra L is the superstitious 'L' 😆. Rectifying it!

PR #652, opened 2019-01-09
Refactor and cleanup tests

sabahirfan:
### Change description ###
This PR cleans up few unit tests and refactors the code. Below are the items cleaned.

1. In `ClaimantResponseMapperTest.java`, these kind of conditions in tests are wrong, because if the condition is false assertion will not be called at all. And test may pass incorrectly.

 ```
if (claim.getClaimantResponse().isPresent()) {
            assertThat((ResponseAcceptation) claim.getClaimantResponse().get()).isEqualTo(ccdResponse);
        }
```
2. In `CourtDeterminationMapper.java` unnecessary `null` check.
```
if (ccdCourtDetermination.getRejectionReason() != null) {
            builder.rejectionReason(ccdCourtDetermination.getRejectionReason());
        }
```
3. Unnecessary unwrapping of `Optionals`
e.g. 
```
assertNotNull(claim.getClaimantRespondedAt().orElse(null));
```

- [ ] Yes
- [x] No

### DIFF
--- a/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/claimantresponse/CourtDeterminationMapper.java
+++ b/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/claimantresponse/CourtDeterminationMapper.java
@@ -37,10 +37,8 @@ public class CourtDeterminationMapper implements Mapper<CCDCourtDetermination, C
             .courtDecision(paymentIntentionMapper.from(ccdCourtDetermination.getCourtDecision()))
             .disposableIncome(ccdCourtDetermination.getDisposableIncome())
             .courtPaymentIntention(paymentIntentionMapper.from(ccdCourtDetermination.getCourtIntention()))
-            .decisionType(ccdCourtDetermination.getDecisionType());
-        if (ccdCourtDetermination.getRejectionReason() != null) {
-            builder.rejectionReason(ccdCourtDetermination.getRejectionReason());
-        }
+            .decisionType(ccdCourtDetermination.getDecisionType())
+            .rejectionReason(ccdCourtDetermination.getRejectionReason());
         return builder.build();
     }
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/claimantresponse/CourtDeterminationMapper.java [hunk 1]
```diff
@@ -37,10 +37,8 @@ public CourtDetermination from(CCDCourtDetermination ccdCourtDetermination) {
             .courtDecision(paymentIntentionMapper.from(ccdCourtDetermination.getCourtDecision()))
             .disposableIncome(ccdCourtDetermination.getDisposableIncome())
             .courtPaymentIntention(paymentIntentionMapper.from(ccdCourtDetermination.getCourtIntention()))
-            .decisionType(ccdCourtDetermination.getDecisionType());
-        if (ccdCourtDetermination.getRejectionReason() != null) {
-            builder.rejectionReason(ccdCourtDetermination.getRejectionReason());
-        }
+            .decisionType(ccdCourtDetermination.getDecisionType())
+            .rejectionReason(ccdCourtDetermination.getRejectionReason());
```
KiranVarma02: rejection reason doesn't exist in most cases, that's the reason we were doing null check before i guess. Is it ok to remove it?

dharmendrak: Kiran it would be set null in this case which is also by default initialised as null, so it doesn't changes anything, good point of using builder and Optional in domain

KiranVarma02: thanks for explaining guys 👍

sabahirfan: You are welcome. 👍

PR #683, opened 2019-01-22
ROC-5104 Proceed with claim creation events when external ID already found in claim store

andrewwa-kainos:
https://tools.hmcts.net/jira/browse/ROC-5104

### Change description ###
When saving a claim whose external ID is already found in the claim-store, raise a new app insights event (`Claim attempt - Duplicate` with the external ID) and proceed with the claim creation event as usual.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
@@ -206,7 +206,7 @@ public class ClaimService {
         } catch (ConflictException e) {
             appInsights.trackEvent(AppInsightsEvent.CLAIM_ATTEMPT_DUPLICATE, CLAIM_EXTERNAL_ID, externalId);
             issuedClaim = caseRepository.getClaimByExternalId(externalId, authorisation)
-                .orElseThrow(RuntimeException::new);
+                .orElseThrow(() -> new NotFoundException("Could not find claim with external ID '" + externalId + "'"));
         }
 
         eventProducer.createClaimIssuedEvent(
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -168,36 +171,43 @@ public Claim saveClaim(
...
+                .letterHolderId(letterHolderId.orElse(null))
+                .features(features)
+                .build();
 
-        Claim issuedClaim = caseRepository.saveClaim(authorisation, claim);
+            issuedClaim = caseRepository.saveClaim(authorisation, claim);
+        } catch (ConflictException e) {
+            appInsights.trackEvent(AppInsightsEvent.CLAIM_ATTEMPT_DUPLICATE, CLAIM_EXTERNAL_ID, externalId);
+            issuedClaim = caseRepository.getClaimByExternalId(externalId, authorisation)
+                .orElseThrow(RuntimeException::new);
```
sabahirfan: Prefer not to throw a generic exception, how about `NotFoundException` instead of `RuntimeException`?

andrewwa-kainos: Yes, good spot, but not sure which exception is most appropriate. In terms of business logic this shouldn't be possible - only if the `caseRepository` somehow lost the claim between calling `getOnHoldIdByExternalId` and here. I'll go with `NotFoundException` for now.

andrewwa-kainos: Wouldn't a `NotFoundException` bubble back as a 404? I think a 500 would be more appropriate for the user

sabahirfan: We don't display these HTTP errors to users, we have generic error page `error.njk` in prod. We will see this in our logs only.

andrewwa-kainos:
> "errorMessage":"Due to an upgrade, you need first to re-analyze the target branch 'master' before analyzing this pull request."

PR #691, opened 2019-01-23
ROC-4131-enabled-dm-for-local-and-other-env

srineelesh:

https://tools.hmcts.net/jira/browse/ROC-4131

### Change description ###
re-enabled dm for local and other environments . 
using the new client DM jar

```
[ X] Yes

```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -14,6 +14,8 @@ locals {
 
   ccdCnpUrl = "http://ccd-data-store-api-${local.local_env}.service.${local.local_ase}.internal"
 
+  document_management_url = "https://dm-store-${local.local_env}.service.${local.local_ase}.internal"
+
   previewVaultName = "${var.raw_product}-aat"
   nonPreviewVaultName = "${var.raw_product}-${var.env}"
   vaultName = "${(var.env == "preview" || var.env == "spreview") ? local.previewVaultName : local.nonPreviewVaultName}"
@@ -192,5 +194,9 @@ module "claim-store-api" {
     ASYNC_MAX_THREADPOOL_SIZE = 50
 
     ROOT_APPENDER = "CMC"
+
+    // document management
+    DOCUMENT_MANAGEMENT_URL = "${local.document_management_url}"
+
   }
 }
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -193,7 +193,7 @@ module "claim-store-api" {
 
     ROOT_APPENDER = "CMC"
 
-    document_management.caseWorkerRoles = "caseworker-cmc"
+    document_management.caseWorkerRole = "caseworker-cmc"
 
   }
 }
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -193,7 +193,7 @@ module "claim-store-api" {
 
     ROOT_APPENDER = "CMC"
 
-    document_management.caseWorkerRole = "caseworker-cmc"
+    DOCUMENT_MANAGEMENT_CASEWORKERROLE = "caseworker-cmc"
 
   }
 }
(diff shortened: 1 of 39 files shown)


infrastructure/main.tf
```diff
@@ -14,6 +14,8 @@ locals {
 
   ccdCnpUrl = "http://ccd-data-store-api-${local.local_env}.service.${local.local_ase}.internal"
 
+  document_management_url = "https://dm-store-${local.local_env}.service.${local.local_ase}.internal"
```
markry-kainos: keep this camelCase to be consistent

infrastructure/main.tf
```diff
@@ -192,5 +194,9 @@ module "claim-store-api" {
     ASYNC_MAX_THREADPOOL_SIZE = 50
 
     ROOT_APPENDER = "CMC"
+
+    // document management
+    DOCUMENT_MANAGEMENT_URL = "${local.document_management_url}"
+
```
markry-kainos: you will also need to define the case worker roles here - similar to others `${var.?}`

src/integrationTest/resources/environment.properties
```diff
@@ -31,6 +31,7 @@ notifications.templates.email.defendantMoreTimeRequested = defendant-more-time-r
 frontend.base.url = https://civil-money-claims.co.uk
 oauth2.client.id = cmc_citizen
 oauth2.client.secret = 123456
+CASE_WORKER_ROLES=caseworker-cmc
```
markry-kainos: suggestion: as this specific to document management - should we define as `document_management.caseWorkerRoles` (seems consistent with others - although the underscore isn't!)

infrastructure/main.tf
markry-kainos: also - i think should be http only? internal services

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
```diff
@@ -34,6 +35,9 @@
     private final AuthTokenGenerator authTokenGenerator;
     private final UserService userService;
 
+    @Value("${CASE_WORKER_ROLES}")
```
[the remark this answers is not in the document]
dharmendrak: it can be as  @Value("${document.management.user-roles}"

infrastructure/main.tf
markry-kainos: sorry @srineelesh !!!! this will need to be moved from here and defined in the per env tfvars file. we only want to turn on for SPROD for now. Define a new variable with default `false` and just define url for SPROD. Once we go to production we can tidy up like this.

src/integrationTest/resources/environment.properties
srineelesh: happy to change it but not sure if this role will never be used for anything else.

infrastructure/main.tf
```diff
@@ -192,5 +194,9 @@ module "claim-store-api" {
     ASYNC_MAX_THREADPOOL_SIZE = 50
 
     ROOT_APPENDER = "CMC"
+
+    // document management
```
sabahirfan: Please remove this comment.

src/integrationTest/resources/environment.properties
sabahirfan: Please use dot(.) instead of underscore(_), in order to stay consistent

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
```diff
@@ -34,6 +35,9 @@
...
+    private String caseWorkerRole;
```
sabahirfan: Please mark this final and initialise in constructor.

PR #699, opened 2019-01-29
ROC-4540-added-app-insights-events-and-unit-tests

srineelesh:

https://tools.hmcts.net/jira/browse/ROC-4540

### Change description ###

Added app insights events where save to and retrieve from DM fails

```
[ X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementBackedDocumentsService.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementBackedDocumentsService.java
@@ -13,7 +13,6 @@ import uk.gov.hmcts.cmc.claimstore.services.ClaimService;
 import uk.gov.hmcts.cmc.domain.models.Claim;
 
 import java.net.URI;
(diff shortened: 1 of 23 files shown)


src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementBackedDocumentsService.java
```diff
@@ -79,18 +90,27 @@ private Claim getClaimByExternalId(String externalId, String authorisation) {
...
+            @Override
+            public PDF getDocument() {
+                return document;
+            }
+        };
+    }
 
-            return document.getBytes();
-        }
+    private interface DocumentDetails {
```
sabahirfan: This interface and anonymous implementation for this is totally unnecessary, we need a data object here which is best achieved by using a `class`

srineelesh: The intention is to avoid creating another class which is highly unlikely to be used outside the context of this class and hence this.

PR #704, opened 2019-02-01
ROC-5093: Execute migration on aat

dharmendrak:
ROC-5093: Execute migration on aat


Jenkinsfile_CNP
```diff
@@ -62,6 +62,10 @@ withPipeline("java", product, component) {
   env.CLAIM_STORE_DB_NAME = 'claimstore'
   env.CLAIM_STORE_DB_USERNAME = 'claimstore'
 
+  after('smokeTest:aat') {
```
markry-kainos: we change this to `after('build')`
then add: `env.CLAIM_STORE_DB_HOST = 'preprod-data-lb.moneyclaim.reform.hmcts.net` (from aat infrastructure)
also password to vault retrieval: `secret('claim-store-db-password', 'CLAIM_STORE_DB_PASSWORD')`

maybe we are missing some other env variables for it to run? but then we can at least run on aat without merging.

PR #712, opened 2019-02-05
ROC-5130 T/A not appearing on PIN letters

ellie-harrison:
https://tools.hmcts.net/jira/browse/ROC-5130

### Change description ###
Add claimant trading as name to pin letter document sent to defendant

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/services/staff/content/DefendantPinLetterContentProvider.java
```diff
@@ -95,4 +96,14 @@ public DefendantPinLetterContentProvider(
         return content;
     }
 
+    private String getPartyNameFor(NamedParty party) {
+        StringBuilder name = new StringBuilder(party.getName());
+
+        if (party instanceof SoleTrader) {
+            (((SoleTrader) party).getBusinessName()).ifPresent(t -> name.append(" T/A ").append(t));
```
andrewwa-kainos: There's an extra pair of parentheses around `getBusinessName()` that could be removed. Not sure using `StringBuilder` is efficient here but not a big deal.

sabahirfan: Prefer `t` to be renamed as `businessName`

PR #720, opened 2019-02-08
ROC-5155: Action product owner suggestion

dharmendrak:


ccd-claim-migration/src/main/java/uk/gov/hmcts/cmc/ccd/migration/services/ClaimMigrator.java
```diff
@@ -176,141 +216,135 @@ private void createCase(
...
-                        && claim.getReDetermination().isPresent()
-                        && claim.getReDetermination().get().getPartyType() == MadeBy.CLAIMANT;
-                case REFER_TO_JUDGE_BY_DEFENDANT:
-                    return claim.getReDeterminationRequestedAt().isPresent()
-                        && claim.getReDetermination().isPresent()
-                        && claim.getReDetermination().get().getPartyType() == MadeBy.DEFENDANT;
-                default:
-                    return false;
-            }
+        switch (event) {
```
markry-kainos: maybe we should include some try/catch handling with logging here in case of any unknown exceptions, NPEs - or will it be handled higher up?

Jenkinsfile_CNP
```diff
@@ -61,16 +61,16 @@ withPipeline("java", product, component) {
...
-//  }
+  env.IDAM_S2S_AUTH_URL = 'http://rpe-service-auth-provider-aat.service.core-compute-aat.internal'
+  env.CORE_CASE_DATA_API_URL = 'http://ccd-data-store-api-aat.service.core-compute-aat.internal'
+  env.CLAIM_STORE_DB_HOST = 'preprod-data-lb.moneyclaim.reform.hmcts.net'
+  env.CLAIM_STORE_DB_PORT = '5432'
+  env.CLAIM_STORE_DB_NAME = 'claimstore'
+  env.CLAIM_STORE_DB_USERNAME = 'claimstore'
+
+  after('smoketest:aat') {
+    sh './gradlew :ccd-claim-migration:assemble migrateClaims -Djava.util.concurrent.ForkJoinPool.common.parallelism=10'
```
sabahirfan: For 100 claims it should be fine. It will depend on how many records we want to test in next dry run. 
If we are choosing 1000 as suggested by Nick, then bump this to 20 or 25.

ccd-claim-migration/src/main/java/uk/gov/hmcts/cmc/ccd/migration/services/ClaimMigrator.java
dharmendrak: yes it is handled higher up

Jenkinsfile_CNP
dharmendrak: we are running on eat with full payload, on prod can try with 1000

karthikkannankainos:
@dharmendrak, @srineelesh   and @markry-kainos  i think we need to discuss how to do this migration of claim and document without everytime altering the build pipeline. Two ways we can think of

1. Externalize these component and use a separate pipeline.
2. Keep these migration start-able through either an endpoint or thru a shell script and then have a separate pipeline to trigger them. 

What we will avoid by doing the above approach.

1. If we want to re-run a migration, we dont need to deploy all the environment and rerun the job wait for it.
2. Also we will not block others waiting for them not to merge any code.
3. We cant pick and choose which environment to run. Only code change in the pipeline can do this.

karthikkannankainos:
@dharmendrak if you think this would bother any scheduled run, let me know I can approve and we can have a separate task to do the change.

PR #735, opened 2019-02-12
ROC-4540-donot-fail-claim-creation-if-dm-upload-fails

srineelesh:
https://tools.hmcts.net/jira/browse/ROC-4540
https://tools.hmcts.net/jira/browse/ROC-5196

### Change description ###

change to not fail claim creation when document management failure occurs

```
[ X] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementBackedDocumentsService.java
```diff
@@ -91,8 +95,8 @@ private Claim getClaimByExternalId(String externalId, String authorisation) {
     }
 
     private DocumentDetails uploadToDocumentManagement(byte[] documentBytes,
```
markry-kainos: v minor - but would prob line break like - this format seems to be project: https://github.com/hmcts/cmc-claim-store/blob/65e606e964aab5c1dabc682c3fcf11aea30ea8cb/src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementBackedDocumentsService.java#L35

PR #738, opened 2019-02-13
ROC-5214 Mapping settlement data need not check that a new offer can be made

andrewwa-kainos:
https://tools.hmcts.net/jira/browse/ROC-5214

### Change description ###
Extracted functional code to new method, had original use new method after its validation, and mapper uses new method instead of the one with the validation.

```
[ ] Yes
[x] No
```


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/offers/SettlementMapper.java
```diff
@@ -38,33 +38,34 @@ public Settlement fromCCDDefendant(CCDDefendant ccdDefendant) {
...
+            .forEach(partyStatement -> addPartyStatement(partyStatement, settlement));
 
         return settlement;
     }
 
     private void addPartyStatement(PartyStatement partyStatement, Settlement settlement) {
-        if (partyStatement.getType().equals(StatementType.OFFER)) {
-            settlement.makeOffer(partyStatement.getOffer().orElse(null),
+        if (StatementType.OFFER.equals(partyStatement.getType())) {
+            settlement.addOffer(
```
andrewwa-kainos: The only functional change in this file is here

ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/offers/SettlementMapper.java
```diff
@@ -38,33 +38,34 @@ public Settlement fromCCDDefendant(CCDDefendant ccdDefendant) {
...
+                partyStatement.getOffer().orElse(null),
                 partyStatement.getMadeBy(),
                 partyStatement.getId()
             );
         }
 
-        if (partyStatement.getType().equals(StatementType.REJECTION)) {
+        if (StatementType.REJECTION.equals(partyStatement.getType())) {
```
dharmendrak: since you are working here, please change .equals to ==

ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/offers/SettlementMapper.java
dharmendrak: I agree with you it doesn't need, same thing can be done in `formaliseSettlement`  method of `FormaliseResponseAcceptanceService` as well. However was it effecting anything or any failures notices

andrewwa-kainos: I'm not familiar with that code, but isn't it actually making the offers and so would need the validation?

sabahirfan: This change looks fine here.

PR #739, opened 2019-02-13
[Do not Merge]ROC-5092- Creating CCD Cases only if payment is successful

karthikkannankainos:
https://tools.hmcts.net/jira/browse/ROC-5092

### Change description ###
Creating CCD Cases only if payment is successful


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CaseEvent.java
```diff
@@ -2,6 +2,7 @@
 
 public enum CaseEvent {
 
+    CREATE_NEW_CASE("IssueClaim"),
```
srineelesh: do we still need SUBMIT_PRE_PAYMENT

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/BaseIntegrationTest.java
```diff
@@ -38,8 +38,7 @@
     protected static final String USER_ID = "1";
     protected static final String JURISDICTION_ID = "CMC";
     protected static final String CASE_TYPE_ID = "MoneyClaimCase";
-    protected static final String SUBMIT_CLAIM_EVENT = "submitClaimEvent";
-    protected static final String SUBMIT_POST_PAYMENT = "SubmitPostPayment";
+    protected static final String CREATE_NEW_CASE = "IssueClaim";
```
srineelesh: may be we can remove SUBMIT_PRE_PAYMENT if it's not needed

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -179,17 +177,17 @@ public Claim saveClaim(
...
-
-            LocalDateTime now = LocalDateTimeFactory.nowInLocalZone();
+            caseRepository.getClaimByExternalId(externalId, authorisation).ifPresent(claim -> {
+                throw new ConflictException(
+                    String.format("Claim already exist with same external reference as %s", externalId));
+            });
 
             Optional<String> letterHolderId = pinResponse.map(GeneratePinResponse::getUserId);
-            LocalDate issuedOn = issueDateCalculator.calculateIssueDay(now);
+            LocalDate issuedOn = issueDateCalculator.calculateIssueDay(LocalDateTimeFactory.nowInLocalZone());
```
srineelesh: can use a static import here for LocalDateTimeFactory.nowInLocalZone()

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -179,17 +177,17 @@ public Claim saveClaim(
 
         Claim issuedClaim;
         try {
-            Long prePaymentClaimId = caseRepository.getOnHoldIdByExternalId(externalId, authorisation);
-
-            LocalDateTime now = LocalDateTimeFactory.nowInLocalZone();
+            caseRepository.getClaimByExternalId(externalId, authorisation).ifPresent(claim -> {
```
dharmendrak: why do we need this? CCD will throw error if case is already created  (Please test locally ) as well as claim store throws conflict exception on save

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/CoreCaseDataService.java
```diff
@@ -188,7 +143,7 @@ public Claim submitPostPayment(String authorisation, Claim claim) {
                 String.format(
                     CCD_UPDATE_FAILURE_MESSAGE,
```
dharmendrak: this need to be CREATE failure

ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CaseEvent.java
dharmendrak: yes, as it is used by migration script, will be removed in another migration PR

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/BaseIntegrationTest.java
karthikkannankainos: Done

ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CaseEvent.java
karthikkannankainos: @dharmendrak will fix migration.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/SaveClaimWithCoreCaseDataStoreTest.java
```diff
@@ -202,7 +193,7 @@ public void shouldFailIssuingClaimEvenWhenCCDStoreFailsToStartEvent() throws Exc
             .andReturn();
 
         assertThat(result.getResolvedException().getMessage())
-            .isEqualTo("Failed updating claim in CCD store for case id 000LR001 on event SUBMIT_POST_PAYMENT");
+            .isEqualTo("Failed updating claim in CCD store for case id 000LR001 on event CREATE_NEW_CASE");
```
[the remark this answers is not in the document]
karthikkannankainos: Good catch. Fixed it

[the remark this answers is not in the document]
karthikkannankainos: The service was using update for post payment initially. Now changed to create in service.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/SaveClaimWithCoreCaseDataStoreTest.java
```diff
@@ -250,21 +239,6 @@ public void shouldIssueClaimEvenWhenCCDStoreFailsToSubmitEvent() throws Exceptio
...
-        given(serviceAuthorisationApi.serviceToken(anyMap())).willThrow(FeignException.class);
-
-        MvcResult result = makeIssueClaimRequest(claimData, SOLICITOR_AUTHORISATION_TOKEN)
-            .andExpect(status().isNotFound())
-            .andReturn();
-
-        assertThat(result.getResolvedException().getMessage())
-            .isEqualTo(String.format("Case %s not found.", claimData.getExternalId().toString()));
-
+            .isEqualTo("Failed updating claim in CCD store for case id 000LR001 on event CREATE_NEW_CASE");
```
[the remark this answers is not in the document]
karthikkannankainos: used previously available mesage "failed storing claim"

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
karthikkannankainos: Currently, we are using this to make sure that there is no claim already created in claim store. This is replacement of previously existing check if any case is already on onhold status. Instead of check on hold, currently we are checking if there is any case there with same external reference id.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
karthikkannankainos: Done

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -179,17 +177,17 @@ public Claim saveClaim(
...
+                throw new ConflictException(
```
sabahirfan: We are throwing this exception here and catching straight away in the same method, and then calling `caseRepository.getClaimByExternalId(externalId, authorisation)` in catch block few lines down. This is not right.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
sabahirfan: In my opinion calling getByExternalId before first save to throw duplicate error, looks unnecessary. 
The underlying repository should throw duplicate error based on the unique index on external id. If there is no unique index on claim number or UUID we should request CCD team to add this one for our claims. 

Note: This additional call will have impact on performance which is already not good, so need to make a call what problem this call is going to solve at the cost of performance.

karthikkannankainos: @sabahirfan there is no implementation of CCD to enforce uniqueness on a field of claim. So only way is to do defensive check.

karthikkannankainos: Also the above contention is claimrepository, which will go away when we move to CCD as main.

andrewwa-kainos: I'd added the `try...catch` block to ensure that the claim creation notifications were sent in a particular communication failure scenario. Whatever changes you make here, just ensure that the event is produced even if the claim already exists.

karthikkannankainos: good one @andrewwa-kainos . Will ensure. I am not changing the flow.

sabahirfan: @andrewwa-kainos We are handling the CONFLICT exception in `front-end`, I am not sure why we need to suppress this exception in backend. Please have a look at:
https://github.com/hmcts/cmc-citizen-frontend/blob/5810c52e916ed4ed3a8834633c88d169a42c5191/src/main/app/claims/claimStoreClient.ts#L77

andrewwa-kainos: Because we still need to proceed with the claim issued event handling.

sabahirfan: That event must have already handled in the first successful save operation. As we rollback the transaction if any of those handlers fail.

karthikkannankainos: Either the Transaction scope is not properly done or we should check if the claim already exist before even taking it to Payment page from front end.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
sabahirfan: Will there be any change in data for the claim between line:201 and line:226? Both are returning the same claim information.

The second call in catch block to get the claim will return exactly the same claim object.

karthikkannankainos: Throwing conflictexception is to explicitly make sure by coding that its unfavourable situation. Handling how you want should be in catch block where we fetch the claim again and handle the exception. The catch block should not get information or use the information from what it has got from Try block.  But I agreed to change it as what @sabahirfan , @dharmendrak  and @andrewwa-kainos has agreed. 

There will be a separate PR to do the cleanup.

sabahirfan: The additional unnecessary get call will decrease the performance while we are on CCD.

karthikkannankainos: The point is even the conflictexception should not occur. These check should be even in controller and throw exception rather than coming till repository. This is leak in transaction. In CCD this wont even happen.

PR #743, opened 2019-02-14
ROC 5194 notify email for defendant on claimant response rejection of states paid or part admission

tharacka:
https://tools.hmcts.net/jira/browse/ROC-5194

### Change description ###
Add a notification email for when claimant rejects a states paid or part admission

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/config/properties/notifications/EmailTemplates.java
```diff
@@ -111,4 +111,7 @@
 
     @NotEmpty
     private String defendantAdmissionResponseToClaimant;
+
+    @NotEmpty
+    private String claimantRejectedPartAdmitOrStatesPaidEmailToDefendant;
```
sabahirfan: As there is no different email template for these, do we really need to say what type of defendant's response claimant is rejecting in name of this field? May be just  `claimantRejectedEmailToDefendant`

src/main/java/uk/gov/hmcts/cmc/claimstore/services/notifications/NotificationToDefendantService.java
```diff
@@ -48,6 +51,25 @@ public void notifyDefendant(Claim claim) {
         );
     }
 
+    public void notifyDefendantOfRejection(Claim claim) {
+        Response response = claim.getResponse().orElseThrow(IllegalArgumentException::new);
+
+        if (isResponseStatesPaid(response) || response.getResponseType() == ResponseType.PART_ADMISSION) {
```
sabahirfan: You are adding two level of filtering to choose ` notifyDefendant(claim);`  once in EventListener `sendNotificationToDefendant(ClaimantResponseEvent event)` and then here, Please move this condition in event listener if at all needed. 

This condition might be un-necessary as well, please verify.

src/main/java/uk/gov/hmcts/cmc/claimstore/config/properties/notifications/EmailTemplates.java
tharacka: As the wording on the GOV Notify template is very specific I have done this. The fear is that if they change the template to be more specific to States Paid/Part Admit as opposed to the general format it is at the moment. I agree with 'claimantRejectedEmailToDefendant' if the template wasn't specific.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/notifications/NotificationToDefendantService.java
singh-surjeet: I agree
AC states rejection of part admission or states paid. Please note even full defence states paid less than claim amount is sent to claim-store as part-admission states paid hence all of the logic can be put at one place.

PR #770, opened 2019-02-22
ROC-3960-store-claimant-copy-claim-form

srineelesh:

https://tools.hmcts.net/jira/browse/ROC-3960

### Change description ###

store the claim issue receipt \ claimant copy in DM 

```
[X ] No
```

### DIFF
--- a/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
+++ b/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
@@ -26,7 +26,7 @@ import static uk.gov.hmcts.cmc.domain.utils.ToStringStyle.ourStyle;
     value = {"totalAmountTillToday", "totalAmountTillDateOfIssue",
         "amountWithInterestUntilIssueDate", "totalInterest",
         "serviceDate", "amountWithInterest", "directionsQuestionnaireDeadline",
-        "sealedClaimDocument"},
+        "claimDocument"},
     allowGetters = true
 )
 @Getter
--- a/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
+++ b/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.cmc.domain.models;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.Builder;
@@ -25,8 +26,7 @@ import static uk.gov.hmcts.cmc.domain.utils.ToStringStyle.ourStyle;
 @JsonIgnoreProperties(
     value = {"totalAmountTillToday", "totalAmountTillDateOfIssue",
         "amountWithInterestUntilIssueDate", "totalInterest",
-        "serviceDate", "amountWithInterest", "directionsQuestionnaireDeadline",
-        "claimDocument"},
+        "serviceDate", "amountWithInterest", "directionsQuestionnaireDeadline"},
     allowGetters = true
 )
 @Getter
@@ -131,6 +131,7 @@ public class Claim {
         return Optional.ofNullable(settlement);
     }
 
+    @JsonIgnore
     public Optional<URI> getClaimDocument(ClaimDocumentType claimDocumentType) {
         if (claimDocumentCollection == null) {
             return Optional.empty();
--- a/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
+++ b/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
@@ -19,7 +19,6 @@ import java.time.LocalDateTime;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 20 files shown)


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
```diff
@@ -19,15 +19,14 @@
...
-import static uk.gov.hmcts.cmc.domain.models.ClaimDocumentType.SEALED_CLAIM;
 import static uk.gov.hmcts.cmc.domain.utils.ToStringStyle.ourStyle;
 
 // Create these fields in JSON when serialize Java object, ignore them when deserialize.
 @JsonIgnoreProperties(
     value = {"totalAmountTillToday", "totalAmountTillDateOfIssue",
         "amountWithInterestUntilIssueDate", "totalInterest",
         "serviceDate", "amountWithInterest", "directionsQuestionnaireDeadline",
-        "sealedClaimDocument"},
+        "claimDocument"},
```
sabahirfan: I think this needs to be ignored in both serialization and deserialization. It's no more simple getter now, it expects one argument to be passed. It will always be  `Optional.empty()` in JSON when serialize.

sabahirfan: Also, please verify if `sealedClaimDocument` URI is being used in any place i.e. front-end/s then serialize that URI.

srineelesh: did a search and I don't see any refrences to the sealedClaimDocumeUri in the front end.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/SaveClaimWithDocumentManagementTest.java
```diff
@@ -88,15 +91,40 @@ private void assertSealedClaimIsUploadedIntoDocumentManagementStore(
...
+    public void shouldLinkDefendantPinLetterAfterUpload() throws Exception {
+        assertSealedClaimIsLinked(SampleClaimData.submittedByClaimant(),
+            AUTHORISATION_TOKEN,
+            DEFENDANT_PIN_LETTER,
+            "000MC001-defendant-pin-letter.pdf");
+    }
+
+    @Test
+    public void shouldLinkClaimIssueReceiptClaimAfterUpload() throws Exception {
+        assertSealedClaimIsLinked(SampleClaimData.submittedByClaimant(),
```
karthikkannankainos: assertDocumentIsLinked? because common method and we are using this method to check claim_issue_receipt is also linked

src/main/java/uk/gov/hmcts/cmc/claimstore/rpa/ClaimIssuedNotificationService.java
```diff
@@ -60,7 +61,7 @@ private EmailData prepareEmailData(Claim claim, List<PDF> documents) {
         EmailContent content = emailContentProvider.createContent(claim);
 
         EmailAttachment sealedClaimPdfAttachment = documents.stream()
-            .filter(document -> document.getFilename().contains("claim-form"))
+            .filter(document -> document.getClaimDocumentType() == SEALED_CLAIM)
```
karthikkannankainos: Suggestion - can we use equals() instead of ==. But Enum can be compared in both ways.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementBackedDocumentsService.java
```diff
@@ -98,6 +97,28 @@ private Claim getClaimByExternalId(String externalId, String authorisation) {
...
+                return documentManagementService.downloadDocument(authorisation, documentSelfPath, baseFileName);
+            } else {
+                PDF document = new PDF(baseFileName, pdfService.createPdf(claim), claimDocumentType);
+                uploadToDocumentManagement(document,
+                    authorisation,
+                    claim);
+                return document.getBytes();
+            }
+        } catch (Exception ex) {
+            return pdfService.createPdf(claim);
```
karthikkannankainos: can we isolate the pdfgeneration exception and handle it differently? Because if the pdfService.createPdf failed, then handling it with same call wont do any better. Just suggestion

srineelesh: basically what ever the exception.. we want to return the document to the user..

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementBackedDocumentsService.java
```diff
@@ -98,6 +97,28 @@ private Claim getClaimByExternalId(String externalId, String authorisation) {
...
+        }
```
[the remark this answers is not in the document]
srineelesh: app insights is already triggered inside DocumentManagementService..

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementBackedDocumentsService.java
karthikkannankainos: but if we have exception from pdfservice also we attempt to call the same. Can we have two catch statements is my question.

PR #773, opened 2019-02-25
ROC-3958-Store-defendant-response-documents-in-DM

srineelesh:

https://tools.hmcts.net/jira/browse/ROC-3958

### Change description ###
Store defendant response documents in DM

```
[X ] Yes
```


src/test/java/uk/gov/hmcts/cmc/claimstore/controllers/SupportControllerTest.java
```diff
@@ -111,4 +122,21 @@ public void shouldResendRPANotifications() {
...
+    @Test
+    public void shouldTriggerUploadDocumentWhenDefendantResponseSubmittedInvoked() {
+        when(claimService.getClaimByReferenceAnonymous(eq("000CM001"))).thenReturn(Optional.of(sampleClaim));
+        controller.resendStaffNotifications("000CM001", "response-submitted", AUTHORISATION);
+        verify(defendantResponseCitizenNotificationsHandler).uploadDocument(any(DefendantResponseEvent.class));
+    }
+
+    @Test
+    public void shouldThrowExceptionIfDefendantResponseSubmittedWhenNoDefendantResponse() {
+        exceptionRule.expect(ConflictException.class);
```
sabahirfan: Throwing `ConflictException` when response does not exist is misleading. Prefer to use `NotFoundException` or `ResponseNotFoundException` here.

src/test/java/uk/gov/hmcts/cmc/claimstore/controllers/SupportControllerTest.java
```diff
@@ -111,4 +122,21 @@ public void shouldResendRPANotifications() {
...
+        exceptionRule.expectMessage("Claim 000CM001 does not have associated response");
+        when(claimService.getClaimByReferenceAnonymous(eq("000CM001")))
+            .thenReturn(Optional.of(SampleClaim.withNoResponse()));
+        controller.resendStaffNotifications("000CM001", "response-submitted", AUTHORISATION);
```
sabahirfan: Minor: `000CM001` can be extracted as const

src/test/java/uk/gov/hmcts/cmc/claimstore/events/claim/DocumentGeneratorTest.java
```diff
@@ -81,4 +87,28 @@ public void shouldPublishDocumentReadyToPrintEventHavingLetterDocumentBeforeClai
         verify(publisher)
             .publishEvent(new DocumentReadyToPrintEvent(claim, defendantLetterDocument, sealedClaimDocument));
     }
+
+    @Test
+    public void shouldTriggerDocumentUploadEventForNonRepresentedClaim() {
```
sabahirfan: Prefer to call it `....CitizenClaim` instead of `...NonRepresentedClaim`

src/main/java/uk/gov/hmcts/cmc/claimstore/events/response/DefendantResponseCitizenNotificationsHandler.java
```diff
@@ -51,4 +63,19 @@ public void notifyClaimantResponse(DefendantResponseEvent event) {
             referenceForClaimant(claim.getReferenceNumber())
         );
     }
+
+    @EventListener
+    public void uploadDocument(DefendantResponseEvent event) {
```
sabahirfan: This method does not belong in this class. This is a citizen notifications handler.

src/test/java/uk/gov/hmcts/cmc/claimstore/controllers/SupportControllerTest.java
srineelesh: I have not introduced this new exception it, it was already being thrown. I am ok to change it but personally I don't see any difference. Please note : there is no "ResponseNotFoundException" as you suggeting that I should create a new class ?

src/test/java/uk/gov/hmcts/cmc/claimstore/events/claim/DocumentGeneratorTest.java
srineelesh: NonRepresentedClaim makes sense to me and besides it is being used in several places already in the code.

src/main/java/uk/gov/hmcts/cmc/claimstore/events/response/DefendantResponseCitizenNotificationsHandler.java
srineelesh: There is no clear indication as to where this method must ideally belong. Can you please elaborate why it should not be here ?

sabahirfan: The class name `DefendantResponseCitizenNotificationsHandler` suggests its a notification handler for citizens. But this method has nothing to do with citizen notification. This method is generating a document and then firing an event for document upload.

src/test/java/uk/gov/hmcts/cmc/claimstore/controllers/SupportControllerTest.java
karthikkannankainos: ResourceNotFound should be thrown only when a Case is missing in REST, the Resource you are referring. Here what is happening is we are trying to make an action on a resource where the state is conflicting what action you are doing. So ConflictException is correct. 

If we are trying to get /case/response when a response is not there, then we throw ResourceNotFound 404

src/main/java/uk/gov/hmcts/cmc/claimstore/events/DocumentUploadEvent.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.Value;
+import uk.gov.hmcts.cmc.claimstore.documents.output.PDF;
+import uk.gov.hmcts.cmc.domain.models.Claim;
+
+import java.util.List;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+@Value
+public class DocumentUploadEvent {
```
sabahirfan: This class looks a ditto copy of `DocumentGeneratedEvent`, not sure we need a different class/event here. 

The events flow I can visualise for response document upload is:
`DefendantResponseEvent -> DocumentGeneratedEvent`
Then Handler/EventListener for `DocumentGeneratedEvent` will do the upload of the document.

srineelesh: Yes,  DocumentUploadEvent contains the same information as DocumentGeneratedEvent , but DocumentGeneratedEvent is subscribed to by several listeners and I don't want those listeners to be invoked when the event is  to just upload a document.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
```diff
@@ -154,12 +161,15 @@ private void resendStaffNotificationOnMoreTimeRequested(Claim claim) {
...
-    private void resendStaffNotificationOnDefendantResponseSubmitted(Claim claim) {
+    private void resendStaffNotificationOnDefendantResponseSubmitted(Claim claim, String authorization) {
         if (!claim.getResponse().isPresent()) {
-            throw new ConflictException(CLAIM + claim.getId() + " does not have associated response");
+            throw new ConflictException(CLAIM + claim.getReferenceNumber() + " does not have associated response");
         }
-        DefendantResponseEvent event = new DefendantResponseEvent(claim);
+        DefendantResponseEvent event = new DefendantResponseEvent(claim, authorization);
         defendantResponseStaffNotificationHandler.onDefendantResponseSubmitted(event);
+        if (!claim.getClaimDocument(DEFENDANT_RESPONSE_RECEIPT).isPresent()) {
```
sabahirfan: I don't think we need this here, please remove this block. We are already generating and uploading where we are trying to get the document from doc-store.

srineelesh: This is meant for the support team. I have put this in for now, as safety net because we don't yet have a support controller for DM. I will be refactoring this code soon.

sabahirfan: Please remove this unnecessary "safety net"

dharmendrak: this is not requirement, please stick to story requirements

PR #785, opened 2019-03-05
Change aseName to not use remote state

markry-kainos:
### Change description ###

core-compute-aat remote state nuked by accident. This PR to fix sandbox deployments but should we fix permanently?

```
[ ] Yes
[ x ] No
```


infrastructure/main.tf
```diff
@@ -3,7 +3,7 @@ provider "azurerm" {
 }
 
 locals {
-  aseName = "${data.terraform_remote_state.core_apps_compute.ase_name[0]}"
+  aseName = "core-compute-${var.env}"
```
timja: can you remove the remote state reference in state.tf too?

markry-kainos: forgot about this!

PR #795, opened 2019-03-11
ROC-5168: Remove Pre payment calls and its reference.

karthikkannankainos:
https://tools.hmcts.net/jira/browse/ROC-5168

### Change description ###
Remove the pre payment calls.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/ClaimController.java
```diff
@@ -119,19 +118,12 @@ public Claim requestMoreTimeToRespond(@PathVariable("externalId") String externa
...
     @ApiOperation("Check whether a claim is linked to a defendant")
     public DefendantLinkStatus isDefendantLinked(@PathVariable("caseReference") String caseReference) {
-        Boolean linked = claimService.getClaimByReferenceAnonymous(caseReference)
+        boolean linked = claimService.getClaimByReferenceAnonymous(caseReference)
             .filter(claim -> claim.getDefendantId() != null)
             .isPresent();
         return new DefendantLinkStatus(linked);
     }
 
-    @PostMapping(value = "/{externalId:" + UUID_PATTERN + "}/pre-payment")
```
sabahirfan: Please make sure to remove call to this endpoint in `citizen-frontend` first and merge that in master before merging this PR.
This is the code that needs to be removed in frontend:
https://github.com/hmcts/cmc-citizen-frontend/blob/master/src/main/app/claims/claimStoreClient.ts#L40

karthikkannankainos: Hi @sabahirfan. The https://github.com/hmcts/cmc-citizen-frontend/pull/1430/files has the change already. Will ask Paul to test it first, then merge the frontend first and then backend. Thats the plan already.

sabahirfan: Thanks

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/DBCaseRepository.java
```diff
@@ -55,7 +54,8 @@ public DBCaseRepository(
     }
 
     public Optional<Claim> getClaimByExternalId(String externalId, String authorisation) {
-        return claimRepository.getClaimByExternalId(externalId);
+        return claimRepository.getClaimByExternalId(externalId)
+            .map(fetchedClaim -> fetchedClaim.toBuilder().id(null).build());
```
andrewwa-kainos: Standard practice (in the Java community at large - I've been trying to make it so with us, too!) is to make each stream operation as simple as possible - chaining to build up the more complicated expressions. As always, judgment can be made on readability. This one would be:
```
return claimRepository.getClaimByExternalId(externalId)
    .map(Claim::toBuilder)
    .map(builder -> builder.id(null))
    .map(ClaimBuilder::build);
```

andrewwa-kainos: Standard practice (in the Java community at large - I've been trying to make it so with us, too!) is to make each stream operation as simple as possible - chaining to build up the more complicated expressions. As always, judgment can be made on readability. This one would be:
```
return claimRepository.getClaimByExternalId(externalId)
    .map(Claim::toBuilder)
    .map(builder -> builder.id(null))
    .map(ClaimBuilder::build);
```

dharmendrak: shouldn't pass null in Id, would be better to look for other option?

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/DBCaseRepository.java
```diff
@@ -209,6 +201,7 @@ public Claim saveClaim(User user, Claim claim) {
 
         return claimRepository
             .getClaimByExternalId(claim.getExternalId())
+            .map(fetchedClaim -> fetchedClaim.toBuilder().id(null).build())
```
dharmendrak: follow other approach.

sabahirfan: We can consider the option to add the property in our `Claim` object to distinguish the origin of the data store(CCD or CLAIM_STORE), in order to identify this in our front end. We don't need to store this in database/CCD. The relevant repository class can populate it accordingly.

karthikkannankainos: @dharmendrak , @sabahirfan we shouldnt create additional property in claim to distinguish if its CCD or Claimstore. I dont know if the previousServiceReferenceNumber will be useful? Let we have a discussion

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/DBCaseRepository.java
karthikkannankainos: @dharmendrak the DB repository is going to be retired, so we chose to make changes to remove id, which first of all is not used at all. Its best practise not to send the ID to outside your data store, which we are already doing.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/DBCaseRepository.java
sabahirfan: `previousServiceReferenceNumber` will not work as that is not in CMC domain model.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/DBCaseRepository.java
sabahirfan: In order to stop IDs outside, prefer to use Jsonignore on the property, not by injecting null into that field.

karthikkannankainos: @sabahirfan we are not using ID only from Claim-store.  So JsonIgnore wont help much

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/DBCaseRepository.java
karthikkannankainos: @sabahirfan I think the previousServiceReferenceNumber is mapped to referenceNumber rt? I think you did that too.

PR #807, opened 2019-03-15
Update CCD client version

timja:


timja:
Tested on DEMO as PR build doesn't test this

PR #837, opened 2019-03-28
ROC-5327: Raise appinsights for every notification failures

dharmendrak:

https://tools.hmcts.net/jira/browse/ROC-5327

### Change description ###
Raise appinsights for every notification failures

```
[ ] Yes
[ ] No
```


email-client/build.gradle
```diff
@@ -21,7 +21,9 @@ dependencies {
 
   annotationProcessor group: 'org.projectlombok', name: 'lombok', version: '1.18.6'
   compileOnly group: 'org.projectlombok', name: 'lombok', version: '1.18.6'
-  
+  compile group: 'uk.gov.hmcts.reform', name: 'java-logging', version: '3.0.2'
+  compile group: 'uk.gov.hmcts.reform', name: 'java-logging-appinsights', version: '3.0.2'
```
sabahirfan: Can we use the latest version for both of the above please? I think it is `5.0.1`

dharmendrak: done

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -195,56 +195,53 @@ public Claim saveClaim(
...
+
+        Claim claim = Claim.builder()
+            .claimData(claimData)
+            .submitterId(submitterId)
+            .issuedOn(issuedOn)
+            .responseDeadline(responseDeadline)
+            .externalId(externalId)
+            .submitterEmail(submitterEmail)
+            .createdAt(nowInUTC())
+            .letterHolderId(letterHolderId.orElse(null))
```
karthikkannankainos: letterHolderId should not be null. Fail it if its null

PR #839, opened 2019-03-29
ROC-5354 Add served date mapping to CCD

simad:
https://tools.hmcts.net/jira/browse/ROC-5354

### Change description ###

Adding served (service) date to CCD 

- [x] tests have been updated / new tests has been added (if needed)

```
[ ] Yes
[x] No
```


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/defendant/DefendantMapper.java
```diff
@@ -51,6 +51,7 @@ public DefendantMapper(
         requireNonNull(claim, "claim must not be null");
 
         CCDRespondent.CCDRespondentBuilder respondentBuilder = CCDRespondent.builder();
+        respondentBuilder.servedDate(claim.getServiceDate());
```
andrewwa-kainos: Shouldn't there be a corresponding change in `.from()`?

sabahirfan: yes, please add the relevant mapping in from as well.

simad: @andrewwa-kainos \ @sabahirfan the serviced date in the claim-store claim is not stored but in inferred from the issued on (+ 5 days) date, so it will be indirectly mapped upstream by the CaseMapper (which maps the issuedOn date). That's why I haven't introduced it in the from.

I can add `serviceDate` as an explicit field of Claim (and map it in the `from()`) but then we kind risk that issuedOn and serviceDate might go out of sync if somebody changes issuedOn after serviceDate is calculated. What do you think?

simad: Decided to go for the second approach, so that we will use the actual field mapped in CCD when mapping the claim back from CCD 👍

simad:
@andrewwa-kainos / @sabahirfan this PR was approved but i had to add https://github.com/hmcts/cmc-claim-store/blob/4e466a6d15ee48ccc2f335c1c30a664f24aa4b7d/src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java#L221 after being able to run aat tests locally 😬

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Claim.java
```diff
@@ -103,6 +105,7 @@ public Claim(
         this.claimData = claimData;
         this.createdAt = createdAt;
         this.issuedOn = issuedOn;
+        this.serviceDate = serviceDate == null ? issuedOn.plusDays(5) : serviceDate;
```
sabahirfan: This condition looks unnecessary here.

PR #842, opened 2019-04-01
ROC-5457 Calculate claimant response deadline and DQ deadline

andrewwa-kainos:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-5457

### Change description ###
Fix calculation of directions questionnaire deadline
Add storage of calculated claimant response deadline (claim-store only)
CCD out of scope.

```
[ ] Yes
[x] No
```


andrewwa-kainos:
PO feedback is that the claimant only has a deadline for defence responses; any kind of admission will not have the deadline.

src/test/java/uk/gov/hmcts/cmc/claimstore/services/ResponseDeadlineCalculatorTest.java
```diff
@@ -22,6 +22,7 @@
 public class ResponseDeadlineCalculatorTest {
 
     private static final int DAYS_FOR_RESPONSE = 14;
+    private static final int DAYS_FOR_CLAIMANT_RESPONSE = 21;
```
sabahirfan: Can we please use same value in our tests as `claimantResponseDays` in application.yaml to avoid any confusion.

andrewwa-kainos:
Confirmed functional tests pass

src/main/java/uk/gov/hmcts/cmc/claimstore/services/DirectionsQuestionnaireDeadlineCalculator.java
```diff
@@ -30,21 +30,17 @@ public DirectionsQuestionnaireDeadlineCalculator(
     }
 
     public LocalDate calculateDirectionsQuestionnaireDeadlineCalculator(LocalDateTime respondedAt) {
-        return calculateFirstWorkingDayAndAddOffset(respondedAt, serviceDays + timeForResponseInDays);
-    }
-
-    private LocalDate calculateFirstWorkingDayAndAddOffset(LocalDateTime dateTime, int offset) {
-        LocalDate date = dateTime.toLocalDate();
+        LocalDate date = respondedAt.toLocalDate().plusDays(serviceDays + timeForResponseInDays);
```
andrewwa-kainos: Sonar complains about this being two `int`s added to satisfy a `long` parameter. I consider it a FP. If it can't be marked that way in sonarcloud.io then I'll change the fields to `long` instead of casting here.

PR #844, opened 2019-04-02
[WIP]ROC-5122: Mediation CSV

KiranVarma02:

https://tools.hmcts.net/jira/browse/ROC-5122

### Change description ###
Generate Mediation CSV and email it everyday

```
[ ] Yes
[ ] No
```


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/MediationRow.java
```diff
@@ -0,0 +1,60 @@
+package uk.gov.hmcts.cmc.domain.models;
+
+import lombok.Builder;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class MediationRow {
+    public static final String SITE_ID = "4";
```
dharmendrak: please add equalandhasocde annotation for Lombok

KiranVarma02: I have added it. Probably a basic question, is that really necessary since I am never going to check/use if two MediationRow objects are equal?

src/main/java/uk/gov/hmcts/cmc/claimstore/jobs/MediationJobSchedulerService.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.quartz.JobBuilder.newJob;
+import static org.quartz.TriggerBuilder.newTrigger;
+
+@Component
+public class MediationJobSchedulerService {
+
+    public void scheduleMediation() {
+
+        try {
+            Scheduler scheduler = StdSchedulerFactory.getDefaultScheduler();
```
sabahirfan: You don't need to start/stop scheduler explicitly. I hope you want to use the same scheduler instance we are using already. Please have a look at `JobSchedulerService` for reference

KiranVarma02: I saw that, i thought that was used for different purpose...i didn't spend too much time on it yet. I will look in to that when i get back to this task..

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/MediationRow.java
andrewwa-kainos: Just a question of good practice. What it means for an object to be equal to another is part of the class definition.

andrewwa-kainos:
From discussion with @dharmendrak , CCD implementation is out-of-scope for now. When using CCD, the report should simply not be generated/sent even if the MILO configuration is defined.

PR #849, opened 2019-04-03
Use IDAM PR Helm chart for whitelisting urls

markry-kainos:
### Change description ###

Testing a cleaner way to manage SIDAM whitelisting for PR URLs. This will add on install and remove on delete (merge).

```
[ ] Yes
[x] No
```


markry-kainos:
This works - tested and checked with Nick in SIDAM team that was actually cleaned up on delete.

PR #868, opened 2019-04-14
ROC-5551: Save PDF file size in Claim store DB

srineelesh:

https://tools.hmcts.net/jira/browse/ROC-5551?src=confmacro

### Change description ###

When a document is uploaded in dm-store, the document size is retrieved and saved in claim-store DB

```
[ X] No
```


andrewwa-kainos:
I can't see where the data is actually stored in the DB. Can you explain it please?

srineelesh:
> I can't see where the data is actually stored in the DB. Can you explain it please?

in DocumentManagementBackedDocumentsService , line 181

PR #887, opened 2019-04-26
ROC-5552 save pdf file size

srineelesh:
https://tools.hmcts.net/jira/browse/ROC-5552

### Change description ###

Save PDF file size in CCD

```
[ X] No
```


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CCDClaimDocument.java
```diff
@@ -15,4 +15,5 @@
     private LocalDateTime authoredDatetime;
     private LocalDateTime createdDatetime;
     private String createdBy;
+    private long size;
```
johnhosolirius: Really minor question. Can this be an integer instead of a long?

PR #888, opened 2019-04-29
ROC-5576 Letter holder not able to access document

srineelesh:

https://tools.hmcts.net/jira/browse/ROC-5576

### Change description ###

Change to use caseworker-cmc role to access documents in DM  .

```
[ X] Yes
```


src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
```diff
@@ -93,7 +103,7 @@ public ClaimDocument uploadDocument(String authorisation, PDF pdf) {
                 .createdBy(OCMC)
                 .build();
         } catch (Exception ex) {
-            throw new DocumentManagementException(String.format("Unable to upload document %s to document management",
+            throw new DocumentManagementException(String.format("Unable to upload document %s to document management.",
```
dharmendrak: would be better to pass `ex` directly as DocumentManagementException(ex). as @Recover logging confirms that it has finally failed after retry.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
```diff
@@ -105,12 +115,7 @@ public ClaimDocument logUploadDocumentFailure(
         PDF pdf
     ) {
         String filename = pdf.getFilename();
-        String errorMessage = String.format(
-            "After retry : unable to upload document %s into document management due to %s",
-            filename, exception.getMessage()
-        );
-
-        logger.info(errorMessage, exception);
+        logger.info(exception.getMessage() + " " + exception.getCause(), exception);
```
dharmendrak: would be better if we keep this log as this confirms retry has been done

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
```diff
@@ -140,7 +144,7 @@ public ClaimDocument logUploadDocumentFailure(
             return resource.getByteArray();
         } catch (Exception ex) {
             throw new DocumentManagementException(
-                String.format("Unable to download document %s from document management", baseFileName), ex);
+                String.format("Unable to download document %s from document management.", baseFileName), ex);
```
dharmendrak: would be better to pass ex directly as DocumentManagementException(ex). as @recover logging confirms that it has finally failed after retry.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
```diff
@@ -152,13 +156,7 @@ public ClaimDocument logUploadDocumentFailure(
...
     ) {
         String filename = baseFileName + ".pdf";
-
-        String errorMessage = String.format(
-            "Failure: unable to download document %s from document management due to %s",
-            filename, exception.getMessage()
-        );
-
-        logger.info(errorMessage, exception);
+        logger.warn(exception.getMessage() + " " + exception.getCause(), exception);
```
dharmendrak: would be better if we keep this log as this confirms retry has been done

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
srineelesh: I think its best to throw a specific custom checked exception coz the root exception could have been anything.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
srineelesh: Not sure what exactly you mean but the message is being logged in the retry block so I don't think we need to specify that explicitly. Also, The previous log message was logging  redundant message.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
srineelesh: I think its best to throw a specific custom checked exception coz the root exception could have been anything.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
srineelesh: same as above

src/main/java/uk/gov/hmcts/cmc/claimstore/services/document/DocumentManagementService.java
markry-kainos: should it be `info` here?

PR #889, opened 2019-04-29
Fixing upload documents in parallel mode

johnhosolirius:

### Change description ###

```
[ ] Yes
[ X] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/events/ccd/CCDCaseHandler.java
```diff
@@ -49,7 +48,7 @@ public CCDCaseHandler(
         this.userService = userService;
     }
 
-    @TransactionalEventListener
```
johnhosolirius: The async nature of these calls means that there's no guarantee that the save does not happen before the document upload. This is one way around it. The other option is to tell Spring give priority to this call using the Order annotation and give it the highest priority.

dependency-check-suppressions.xml
```diff
@@ -151,4 +151,11 @@
         <gav regex="true">.*</gav>
         <cve>CVE-2019-3795</cve>
     </suppress>
+    <suppress>
+        <notes><![CDATA[
+			Not applicable - A Remote Code Execution Vulnerability in Apache Tomcat
```
srineelesh: can you please remove this. Silvia has made a fix already .https://github.com/hmcts/cmc-claim-store/pull/890

src/main/java/uk/gov/hmcts/cmc/claimstore/events/ccd/CCDCaseHandler.java
johnhosolirius: The async calls means that there's no guarantee that this call is executed before the document upload event if it's transactional

src/main/java/uk/gov/hmcts/cmc/claimstore/events/ccd/CCDCaseHandler.java
```diff
@@ -148,6 +147,26 @@ public void linkDefendantToClaim(CCDLinkDefendantEvent event) {
...
+                event.getAuthorisation()
+            ).orElseThrow(IllegalStateException::new);
+
+            ccdCaseRepository.saveClaimDocuments(event.getAuthorisation(),
+                ccdClaim.getId(),
+                event.getClaimDocumentCollection(),
+                event.getClaimDocumentType()
+            );
+        } catch (FeignException e) {
+            appInsights.trackEvent(CCD_ASYNC_FAILURE, REFERENCE_NUMBER, event.getClaim().getReferenceNumber());
```
srineelesh: why are we catching only FeignException . what happens if there is any other exception.

johnhosolirius: All the methods in CCDCaseHandler do that, so it follows the pattern. But yeah, it's a bit strange.

srineelesh: This is strange. anyways its just a suggestion. This can be fixed , if need be, in a separate pr

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -238,15 +238,14 @@ public Claim saveClaim(
                 .build();
 
             issuedClaim = caseRepository.saveClaim(user, claim);
+            ccdEventProducer.createCCDClaimIssuedEvent(issuedClaim, user);
```
ckalista: 👍

src/main/java/uk/gov/hmcts/cmc/claimstore/events/ccd/CCDCaseHandler.java
ckalista: Good catch!

PR #891, opened 2019-04-29
Exclude lombok generated code from coverage scan

sabahirfan:
### Change description ###
This PR excludes the lombok generated code from coverage scan.

- [ ] Yes
- [x] No


lombok.config
```diff
@@ -0,0 +1 @@
+lombok.addLombokGeneratedAnnotation = true
```
andrewwa-kainos: From the lombok site:
> We advise against this; JDK9 breaks this annotation, and it's unlikely to ever get fixed.

Is there no longer-term resolution?

sabahirfan: @andrewwa-kainos That comment is for the property `lombok.addJavaxGeneratedAnnotation`

PR #892, opened 2019-04-29
ROC-5598 - Resend Claimant Response Staff Notifications from Support Controller

ellie-harrison:
https://tools.hmcts.net/jira/browse/ROC-5598

### Change description ###
Add new event for resending claimant response staff notifications to support controller.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
```diff
@@ -236,6 +254,33 @@ private void resendClaimsToRPA(List<Claim> claims, String authorisation) {
         }
     }
 
+    private void resendStaffNotificationClaimantResponse(Claim claim) {
+        ClaimantResponse claimantResponse = claim.getClaimantResponse()
+            .orElseThrow(IllegalStateException::new);
+        Response response = claim.getResponse().orElseThrow(IllegalArgumentException::new);
```
andrewwa-kainos: Is there really a difference between a missing claimant response and a missing response? I think they should both be `IllegalArgumentException`s

PR #896, opened 2019-05-01
Fix idam intermittent failures in tests

markry-kainos:
### Change description ###

Testing delay after user creation again...

```
[ ] Yes
[x] No
```


src/aat/java/uk/gov/hmcts/cmc/claimstore/tests/idam/IdamTestService.java
```diff
@@ -96,6 +100,16 @@ public User createDefendant(final String letterHolderId) {
         return userService.authenticateUser(email, password);
     }
 
+    private void createUser(CreateUserRequest createUserRequest) {
+        idamTestApi.createUser(createUserRequest);
+        //recommended delay from SIDAM team to stop intermittent auth failures
+        try {
+            Thread.sleep(1000);
```
sabahirfan: So we still need this delay. IMHO This should be a temporary solution not permanent.  Should we create a story/bug for IDAM? and have a tech debit story in our Jira to link with that?

PR #973, opened 2019-05-28
ROC-5698 Use the full claim amount in CCJ having rejected part-admit

andrewwa-kainos:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-5698

### Change description ###
Only use the part-admit amount in the robotics PDF if the claimant actually accepted that amount.

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/cmc/claimstore/services/staff/content/countycourtjudgment/AmountContentProviderTest.java
```diff
@@ -61,9 +61,9 @@ public void calculateAmountDetailsWithPartAdmissions() {
             .build();
 
         AmountContent amountContent = amountContentProvider.create(claim);
-        assertThat(amountContent.getAdmittedAmount()).isEqualTo("£120.99");
-        assertThat(amountContent.getSubTotalAmount()).isEqualTo("£160.99");
-        assertThat(amountContent.getRemainingAmount()).isEqualTo("£160.99");
+        assertThat(amountContent.getAdmittedAmount()).isEqualTo("£40.99");
```
robertparkinson: why do these need to be changed?

andrewwa-kainos: Good point. It isn't used in the PDF in _this_ case, but I'll make sure it's correct anyway in case it's used in others.

PR #979, opened 2019-05-29
ROC-5752 Include more useful information about claims in the metadata results

andrewwa-kainos:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-5752

### Change description ###
Add more information to the metadata results

```
[ ] Yes
[x] No
```


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/metadata/ClaimantResponseMetadata.java
```diff
@@ -0,0 +1,103 @@
...
+        final Boolean settleForAmount = optionalSettleForAmount.isPresent()
+            ? optionalSettleForAmount.filter(isEqual(YesNoOption.YES)).isPresent()
+            : null;
+        final Optional<BigDecimal> optionalAmountPaid = claimantResponse.getAmountPaid();
+        final Boolean fullPaymentReceived = optionalAmountPaid.isPresent()
+            ? optionalAmountPaid.filter(isEqual(claim.getTotalAmountTillToday())).isPresent()
+            : null;
+        final ClaimantResponseType claimantResponseType = claimantResponse.getType();
+
+        if (claimantResponse instanceof ResponseAcceptation) {
```
sabahirfan: Can we please use `claimantResponse.getType()` in switch instead of `instanceof` with `if`?
We can also extract out the individual methods for each type to reduce the complexity of this method.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/metadata/ClaimantResponseMetadata.java
```diff
@@ -0,0 +1,105 @@
...
+                final Boolean mediation = extractBooleanOrNull(responseRejection::getFreeMediation, YesNoOption.YES);
+                return new ClaimantResponseMetadata(
+                    claimantRespondedAt,
+                    claimantResponseType,
+                    fullPaymentReceived,
+                    settleForAmount,
+                    null,
+                    null,
+                    null,
+                    null,
```
sabahirfan: Rather than passing nulls in constructor, would be nice to use builder in these scenarios

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/metadata/ClaimantResponseMetadata.java
sabahirfan: Thanks 👍

PR #990, opened 2019-06-03
Fix interest end date issue

dharmendrak:
### Change description ###

Fix interest end date issue

```
[ ] Yes
[ ] No
```


ccd-claim-migration/src/main/java/uk/gov/hmcts/cmc/ccd/migration/services/ClaimMigrator.java
```diff
@@ -100,14 +109,26 @@ private void migrateClaims(
...
         notMigratedClaims.parallelStream().forEach(claim -> {
-            migrationHandler.migrateClaim(
-                migratedClaims,
-                failedOnCreateMigrations,
-                failedOnUpdateMigrations,
-                updatedClaims,
-                claim,
-                user
-            );
+            if (fixDataIssues && isSettledOrJudgement(claim)) {
```
simad: Does this mean that claimed which are "fixed" are not migrated? is this what we want? (I kind of lack context about this so it's a n00b question :D)

ccd-claim-migration/src/main/java/uk/gov/hmcts/cmc/ccd/migration/services/DataFixHandler.java
```diff
@@ -0,0 +1,86 @@
...
+        AtomicInteger failedMigrations,
+        Claim claim,
+        CaseDetails caseDetails
+    ) {
+        String referenceNumber = claim.getReferenceNumber();
+        try {
+
+            logger.info("start updating case for: {} for event: {}", referenceNumber, TEST_SUPPORT_UPDATE);
+
+            updateCCDCaseService.updateCase(user, caseDetails.getId(), claim, TEST_SUPPORT_UPDATE);
```
ckalista: We should probably use a new type and call it something that doesn't mention `TEST` which could be misleading for case workers. Will check with Marc what he would prefer it to be called.

ccd-claim-migration/src/main/java/uk/gov/hmcts/cmc/ccd/migration/services/ClaimMigrator.java
```diff
@@ -100,14 +109,26 @@ private void migrateClaims(
...
+                dataFixHandler.fixClaim(migratedClaims, failedOnUpdateMigrations, updatedClaims, claim, user);
+            } else {
```
sabahirfan: This condition have a leakage. This will try to migrate claim if the `isSettledOrJudgement(claim)` returns `false` which is wrong. 

We should not call migrateClaim at all if we have flag `fixDataIssues` as `true`.

ccd-claim-migration/src/main/java/uk/gov/hmcts/cmc/ccd/migration/services/ClaimMigrator.java
ckalista: We are recycling our migration script here Silvia. In theory we should never have to use it to migrate a single claim ever again. @dharmendrak shall we make this a little more robust though to only rely on a single condition `fixDataIssues` - just worried we may otherwise migrate unwanted claims...

ccd-claim-migration/src/main/java/uk/gov/hmcts/cmc/ccd/migration/services/ClaimMigrator.java
```diff
@@ -100,14 +109,26 @@ private void migrateClaims(
...
+                migrationHandler.migrateClaim(
```
karthikkannankainos: Just for this run, can we comment this migrateClaim... ? If by mistake we send a claim which is not in if condition, this will duplicate the claim by migrating again.

ccd-claim-migration/src/main/java/uk/gov/hmcts/cmc/ccd/migration/services/ClaimMigrator.java
dharmendrak: good one

ccd-claim-migration/src/main/java/uk/gov/hmcts/cmc/ccd/migration/services/DataFixHandler.java
```diff
@@ -0,0 +1,86 @@
...
+        AtomicInteger updatedClaims,
+        Claim claim,
+        User user
+    ) {
+        try {
+            Optional<CaseDetails> caseDetails
+                = searchCCDCaseService.getCcdIdByReferenceNumber(user, claim.getReferenceNumber());
+
+            caseDetails.ifPresent(details -> {
+                if (details.getState().equals(ClaimState.OPEN.getValue())) {
```
karthikkannankainos: Are we suppose to fix only the OPEN claims?, better to remove this condition.

ccd-claim-migration/src/main/java/uk/gov/hmcts/cmc/ccd/migration/services/DataFixHandler.java
karthikkannankainos: @ckalista that will mean we have to do a definition release.

ckalista: Yes, unfortunately - let's do it. Marc also wants this new event to be hidden.

PR #995, opened 2019-06-06
ROC-5800:support endpoint to reset claim submission indicators

srineelesh:

https://tools.hmcts.net/jira/browse/ROC-5800

### Change description ###
Adding a new endpoint to allow resetting of the claim submission indicators

```
[ X] Yes
```


src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
```diff
@@ -163,6 +165,25 @@ public void uploadDocumentToDocumentManagement(
...
+    @PostMapping("/claim/{referenceNumber}/claimSubmissionOperationIndicators")
+    @ApiOperation("update the claim submission operation Indicators")
+    public void updateClaimSubmissionIndicators(
+        @PathVariable("referenceNumber") String referenceNumber,
+        @RequestBody ClaimSubmissionOperationIndicators claimSubmissionOperationIndicators,
+        @RequestHeader(value = HttpHeaders.AUTHORIZATION) String authorisation
+    ) {
+        Claim claim = claimService.getClaimByReferenceAnonymous(referenceNumber)
+            .orElseThrow(() -> new NotFoundException(String.format(CLAIM_DOES_NOT_EXIST, referenceNumber)));
+        if (StringUtils.isBlank(authorisation)) {
```
markry-kainos: should we move this up?

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -354,4 +355,13 @@ public void saveReDetermination(
     public void updateClaimState(String authorisation, Claim claim, ClaimState state) {
         caseRepository.updateClaimState(authorisation, claim.getId(), state);
     }
+
+    public void updateClaimSubmissionOperationIndicators(String authorisation,
```
markry-kainos: i see different formatting for this every time 👀 maybe for now we should just be consistent with rest of class and refactor everything together

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -354,4 +355,13 @@ public void saveReDetermination(
...
+                                         Claim claim,
+                                         ClaimSubmissionOperationIndicators claimSubmissionOperationIndicators) {
+        caseRepository.updateClaimSubmissionOperationStatus(authorisation,
```
markry-kainos: formatting here too - just be consistent with class and put a case forward for change 😸

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
karthikkannankainos: @markry-kainos @srineelesh pls let we not refactor rest of class for this change. I think it should be taken as separate PR. Also we havent as team agreed what is standard yet. So just leave as it is

srineelesh: @markry-kainos I have changed the format to keep it consistent with the others. I personally don't prefer this format because the code format is in accordance with checkstyle and intellij. I haven't changed the entire class because as Karthik mentioned I too feel it should be in a separate PR. I think we should decide as a team what format to follow and best is to set it up in cheekstyle.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
srineelesh: done..

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
srineelesh: done.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
markry-kainos: cool - i meant that - keep same and we change big later 😊 👍

ccd-adapter/src/test/java/uk/gov/hmcts/cmc/ccd/util/SampleData.java
```diff
@@ -487,6 +488,7 @@ public static CCDCase getCCDLegalCase() {
             .applicants(applicants)
             .respondents(respondents)
             .claimSubmissionOperationIndicators(getDefaultCCDClaimSubmissionOperationIndicators.get())
+            .state("open")
```
sabahirfan: Prefer to use `ClaimState.OPEN.getValue()` instead of string here

ccd-adapter/src/test/java/uk/gov/hmcts/cmc/ccd/util/SampleData.java
```diff
@@ -113,6 +113,7 @@
         .paymentAmount("400000")
         .paymentReference("RC-1524-6488-1670-7520")
         .claimSubmissionOperationIndicators(getDefaultCCDClaimSubmissionOperationIndicators.get())
+        .state("create")
```
sabahirfan: Prefer to use `ClaimState.CREATE.getValue()` instead of string here

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/SaveClaimWithCoreCaseDataStoreTest.java
```diff
@@ -25,7 +25,7 @@
         "document_management.url=false",
         "feature_toggles.ccd_async_enabled=false",
         "feature_toggles.ccd_enabled=true",
-        "feature_toggles.async_event_operations_enabled=false"
+        "feature_toggles.async_event_operations_enabled=true"
```
sabahirfan: Is this required and related to this ?

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
```diff
@@ -163,6 +167,76 @@ public void uploadDocumentToDocumentManagement(
...
+            claimSubmissionOperationIndicators);
+        claim = claimService.updateClaimSubmissionOperationIndicators(
+            authorisation,
+            claim,
+            claimSubmissionOperationIndicators
+        );
+        triggerAsyncOperation(authorisation, claim);
+    }
+
+    private void validateClaimSubmissionOpsInd(
```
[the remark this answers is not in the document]
sabahirfan: Please move this validation out of the controller, preferably create a custom validator and apply that on the request payload.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseRepository.java
```diff
@@ -212,7 +212,7 @@ public void saveCaseEvent(String authorisation, Claim claim, CaseEvent caseEvent
 
     @Override
     public void updateClaimState(String authorisation, Long claimId, ClaimState state) {
-        if (state == ClaimState.OPEN) {
+        if (state == ClaimState.CREATE) {
```
sabahirfan: Also you don't need this check here as you already have this here (before calling this):
https://github.com/hmcts/cmc-claim-store/pull/995/files#diff-c18491d71318a89d1f0de872c9af15cdR117

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/CoreCaseDataService.java
```diff
@@ -791,7 +792,7 @@ public void saveCaseEvent(String authorisation, Long caseId, CaseEvent caseEvent
             );
 
             CCDCase ccdCase = extractCase(startEventResponse.getCaseDetails());
-
+            ccdCase.setState(null);
```
sabahirfan: Please don't set state to null explicitly.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseRepository.java
karthikkannankainos: The main point of having the check here is to make sure nobody calls this method for state transition other than create -> open. So this check should stay here.

src/main/java/uk/gov/hmcts/cmc/claimstore/events/claim/PostClaimOrchestrationHandler.java
```diff
@@ -134,14 +139,20 @@ public void representativeIssueHandler(RepresentedClaimCreatedEvent event) {
             Function<Claim, Claim> doUploadSealedClaim =
                 c -> uploadSealedClaimOperation.perform(c, authorisation, sealedClaim);
 
-            Supplier<Claim> updatedClaim = () ->
```
sabahirfan: Thank you for clearing this unnecessary Supplier 👍

domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleClaim.java
```diff
@@ -90,8 +89,8 @@
     private ClaimDocumentCollection claimDocumentCollection = new ClaimDocumentCollection();
     private LocalDate claimantResponseDeadline;
     private ClaimState state = null;
-    private static Supplier<ClaimSubmissionOperationIndicators> getDefaultClaimSubmissionOperationIndicators =
```
sabahirfan: Thank you for clearing this unnecessary Supplier 👍

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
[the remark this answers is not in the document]
srineelesh: Please can you point me to existing custom validator which does something like this? I have seen constraint validators in our application which work slightly differently. Not sure if I can use them here.  The validation already has unit tests please have a look at my changes to SupportControllerTest.java .

ccd-adapter/src/test/java/uk/gov/hmcts/cmc/ccd/util/SampleData.java
srineelesh: done

ccd-adapter/src/test/java/uk/gov/hmcts/cmc/ccd/util/SampleData.java
srineelesh: done

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/SaveClaimWithCoreCaseDataStoreTest.java
srineelesh: yeah .. if this is not true then it triggers the old claimissueevent. IMHO where ever applicable we need to turn on feature_toggles.async_event_operations_enabled=true

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseRepository.java
```diff
@@ -212,7 +212,7 @@ public void saveCaseEvent(String authorisation, Claim claim, CaseEvent caseEvent
 
     @Override
     public void updateClaimState(String authorisation, Long claimId, ClaimState state) {
```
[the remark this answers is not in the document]
srineelesh: The name seems ok to me and besides, it was not added as part of this PR. Have changed the name to updateClaimStateToOpen . This implementation is as per the discussion we had with @chrisisk .

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseRepository.java
srineelesh: agree with @karthikkannankainos on this.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/CoreCaseDataService.java
srineelesh: In this particular method I found that we are not using the mappers to map from cmc domain model to ccd model  which is why the case state has to be set to null explicitly. Since we do not maintain the state in ccd definition it throws errors if it is not removed i.e set to null .I am open to any viable suggestions or workarounds.

domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleClaim.java
```diff
@@ -90,8 +89,8 @@
...
-        () -> ClaimSubmissionOperationIndicators.builder().build();
+    private ClaimSubmissionOperationIndicators claimSubmissionOperationIndicators =
```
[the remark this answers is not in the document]
srineelesh: Done. 'You are star' --  @karthikkannankainos thank you Miss !!

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/CoreCaseDataService.java
sabahirfan: We can use `@JsonProperty(access = JsonProperty.Access.WRITE_ONLY)` on the property itself

srineelesh: good one. thanks. Done

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseRepository.java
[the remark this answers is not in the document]
sabahirfan: Thanks

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseRepository.java
sabahirfan: The state machine is controlled by CCD based on the event we fire, So the whole purpose of this method is to fire an event and CCD will change it's state based on the postcondition defined. CCD will check the precondition before applying this ISSUE_CLAIM event.

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
```diff
@@ -163,6 +167,76 @@ public void uploadDocumentToDocumentManagement(
...
+        ClaimSubmissionOperationIndicators existingClaimSubmissionOperationIndicators,
+        ClaimSubmissionOperationIndicators newClaimSubmissionOperationIndicators
+    ) {
+        List<String> invalidIndicators = new ArrayList<>();
+
+        if (existingClaimSubmissionOperationIndicators.getClaimIssueReceiptUpload().equals(NO)
```
sabahirfan: There was a comment which seems lost now. 
It was to move all this validation out of controller, can have a ConstraintValidator which we can apply on the request payload. And it can have it's own unit tests to cover all scenarios.

srineelesh: @sabahirfan yes Indeed and I had replied to the comment as well. I don't know how it got lost. In my reply I had asked , if you can point me to a custom validator which does similar validations. In the application, I see only constraint validators and they work slightly differently. For this particular validation, we need to fetch the claim first whereas I don't see that for other constraint validators. Can you please point me to a sample one where we do this kind of operation. Also the unit tests are already there . Please have a look at SuppotControllerTest.java

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseRepository.java
srineelesh: This means we make another call to CCD which we can otherwise avoid.

dharmendrak: @karthikkannankainos I was told that state here passed here should be OPEN, then we call CLAIM_ISSUE Event, otherwise error, why we are changing to `CREATE`

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
sabahirfan: You can find plenty of examples under this package `uk.gov.hmcts.cmc.domain.constraints`. 
Unit test for the validation in controller test is wrong place.

srineelesh: I have already seen some of the validators in this package.. please can you point me to one where we are fetching the claim and validating against the data in the claim.

sabahirfan: Payload validation should be moved in a constraint, if there is more than that create a Rule class and move that logic there. You can find examples of Rules under `uk.gov.hmcts.cmc.claimstore.rules`

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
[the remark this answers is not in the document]
sabahirfan: Prefer to consider codacy suggestions

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseRepository.java
karthikkannankainos: @dharmendrak the logic has changed here. Please see if the code change applies intact

domain-sample-data/src/main/java/uk/gov/hmcts/cmc/domain/models/sampledata/SampleClaim.java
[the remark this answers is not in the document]
karthikkannankainos: Good that atleast few people understand what is the use of Supplier. Thanks

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimState.java
```diff
@@ -15,4 +15,13 @@
...
         return state;
     }
+
+    public static ClaimState fromValue(String state) {
+        for (ClaimState claimState : ClaimState.values()) {
+            if (claimState.getValue().equalsIgnoreCase(state)) {
+                return claimState;
+            }
+        }
+        throw new IllegalArgumentException(state + " is not a valid state.");
```
andrewwa-kainos: Something like this should be easier to read
```
return Arrays.stream(ClaimState.values())
    .filter(claimState -> claimState.getValue().equalsIgnoreCase(state))
    .findAny()
    .orElseThrow(() -> new IllegalArgumentException(state + " is not a valid state.");
```

src/main/java/uk/gov/hmcts/cmc/claimstore/events/claim/PostClaimOrchestrationHandler.java
```diff
@@ -106,16 +105,22 @@ public void citizenIssueHandler(CitizenClaimCreatedEvent event) {
...
-            claimService.updateClaimState(authorisation, updatedClaim.get(), ClaimState.OPEN);
-
+            updatedClaim.getState()
+                .filter(state -> state == CREATE)
+                .ifPresent(state -> {
+                    claimService.updateClaimState(
+                        authorisation,
+                        updatedClaim,
+                        state);
+                });
```
andrewwa-kainos: `.filter(state -> state == CREATE)` could be simplified to `.filter(Predicate.isEqual(CREATE))`
(static import on `isEqual` to further simplify)
`.ifPresent(...)` body doesn't need braces as it's a single statement.

src/main/java/uk/gov/hmcts/cmc/claimstore/events/claim/PostClaimOrchestrationHandler.java
```diff
@@ -134,14 +139,20 @@ public void representativeIssueHandler(RepresentedClaimCreatedEvent event) {
...
+                .apply(claim);
+
+            updatedClaim.getState()
+                .filter(state -> state == CREATE)
+                .ifPresent(state -> {
+                    claimService.updateClaimState(
+                        authorisation,
+                        updatedClaim,
+                        state);
+                });
```
andrewwa-kainos: `.filter(state -> state == CREATE)` could be simplified to `.filter(Predicate.isEqual(CREATE))`
(static import on `isEqual` to further simplify)
`.ifPresent(...)` body doesn't need braces as it's a single statement.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CaseRepository.java
```diff
@@ -84,7 +84,7 @@ Claim updateClaimSubmissionOperationStatus(
         ClaimSubmissionOperationIndicators indicators,
         CaseEvent caseEvent);
 
-    void updateClaimState(String authorisation, Long claimId, ClaimState state);
+    void updateClaimStateToOpen(String authorisation, Long claimId, ClaimState state);
```
andrewwa-kainos: If `state` is guaranteed to be `OPEN` then we don't need the parameter; otherwise the method name change is inappropriate.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimState.java
srineelesh: I felt going out of the way to create a stream out of an array would be an overkill in this case. Have a look at this https://jaxenter.com/java-performance-tutorial-how-fast-are-the-java-8-streams-118830.html

If you still insist and feel this is necessary, I will change it.

src/main/java/uk/gov/hmcts/cmc/claimstore/events/claim/PostClaimOrchestrationHandler.java
srineelesh: done.

src/main/java/uk/gov/hmcts/cmc/claimstore/events/claim/PostClaimOrchestrationHandler.java
srineelesh: done

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CaseRepository.java
srineelesh: state is not guaranteed to be open. we check that in the implementation. The name change was to make it clear that we are changing the state to Open.

andrewwa-kainos: As discussed offline: now I understand the purpose of this method, I'd like to see the `state` parameter renamed to `currentState` or `previousState` to make it clear that it's not specifying the target state.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimState.java
srineelesh: done

src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
srineelesh: I do not agree with this comment. I see validation being done in other methods in this controller as well.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseRepository.java
srineelesh: As per discussion with the reviewer, I have removed the exception being thrown. Also changed the all invoking code to send the current state of the claim this may cause issues if attempted to run in a non async mode.

src/main/java/uk/gov/hmcts/cmc/claimstore/rules/ClaimSubmissionOperationIndicatorRule.java
```diff
@@ -0,0 +1,132 @@
...
+import javax.validation.constraints.NotNull;
+
+import static uk.gov.hmcts.cmc.domain.models.ClaimDocumentType.CLAIM_ISSUE_RECEIPT;
+import static uk.gov.hmcts.cmc.domain.models.response.YesNoOption.NO;
+import static uk.gov.hmcts.cmc.domain.models.response.YesNoOption.YES;
+
+@Service
+public class ClaimSubmissionOperationIndicatorRule {
+
+    private YesNoOption sealedClaimUpload;
```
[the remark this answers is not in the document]
sabahirfan: Please remove these unused members.

src/main/java/uk/gov/hmcts/cmc/claimstore/rules/ClaimSubmissionOperationIndicatorRule.java
```diff
@@ -0,0 +1,132 @@
...
+    private YesNoOption claimIssueReceiptUpload;
```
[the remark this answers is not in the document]
sabahirfan: Please remove these unused members.

PR #1002, opened 2019-06-12
Handle Null interest type in claim interest

dharmendrak:
### Change description ###

Handle Null interest type in claim interest

```
[ ] Yes
[ ] No
```


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/InterestMapper.java
```diff
@@ -37,8 +39,10 @@ public void to(Interest interest, CCDCase.CCDCaseBuilder builder) {
         interestBreakdownMapper.to(interest.getInterestBreakdown(), builder);
         interestDateMapper.to(interest.getInterestDate(), builder);
 
+        Optional.ofNullable(interest.getType())
```
sabahirfan: As per `uk.gov.hmcts.cmc.domain.models.Interest` this should not be null. 

```
@NotNull
    private final InterestType type;
```
Not sure why we still need to do this?

dharmendrak: I agree on it, but looks like claim store have some data

PR #1018, opened 2019-06-20
ROC-5849: Use binary link from document

dharmendrak:

https://tools.hmcts.net/jira/browse/ROC-5849

### Change description ###

Use binary link from document

```
[ ] Yes
[ ] No
```


ccd-adapter/src/test/java/uk/gov/hmcts/cmc/ccd/mapper/ClaimDocumentMapperTest.java
```diff
@@ -71,9 +72,13 @@ public void mapFrom() {
         ClaimDocument claimDocument = claimDocumentMapper.from(ccdCollectionElement);
         CCDClaimDocument ccdClaimDocument = ccdCollectionElement.getValue();
 
-        assertEquals(ccdClaimDocument.getDocumentName(), claimDocument.getDocumentName());
         assertEquals(URI.create(ccdClaimDocument.getDocumentLink().getDocumentUrl()),
             claimDocument.getDocumentManagementUrl());
+
+        assertEquals(URI.create(ccdClaimDocument.getDocumentLink().getDocumentUrl()),
+            claimDocument.getDocumentManagementUrl());
```
andrewwa-kainos: Duplicates the previous assertion. I guess this is meant to be about the binary URL

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimDocument.java
```diff
@@ -9,6 +9,7 @@
 @Getter
 public class ClaimDocument extends CollectionId {
     private final URI documentManagementUrl;
+    private final URI documentManagementBinaryUrl;
```
sabahirfan: do we need both of these urls? if yes, we should have same in `CCDClaimDocument` as well

sabahirfan: Ignore this, as it's part of `documentLink` property in CCD side.

PR #1019, opened 2019-06-21
ROC-5871: Allow claimant to fetch created claims

sabahirfan:
[ROC-5871](https://tools.hmcts.net/jira/browse/ROC-5871)

### Change description ###
This PR implements the access to cases in created cases for claimant, but for defendant only cases in open state are returned.

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
@@ -285,7 +285,7 @@ public class CCDCaseApi {
 
         User letterHolder = userService.getUser(authorisation);
         CaseDetails caseDetails = readCase(letterHolder, letterHolderCases.get(0));
-        if (CREATE == ClaimState.valueOf(caseDetails.getState())) {
+        if (CREATE == ClaimState.fromValue(caseDetails.getState())) {
             throw new DefendantLinkingException("Claim is not in issued yet, can not link defendant");
         }
         return Optional.of(ccdCaseDataToClaim.extractClaim(caseDetails));
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import static uk.gov.hmcts.cmc.ccd.util.StreamUtil.asStream;
@@ -99,7 +100,7 @@ public class CCDCaseApi {
     public List<Claim> getByDefendantId(String id, String authorisation) {
         User user = userService.getUser(authorisation);
 
-        return asStream(getAllCasesBy(user, ImmutableMap.of()))
+        return asStream(getAllIssuedCasesBy(user, ImmutableMap.of()))
             .filter(claim -> id.equals(claim.getDefendantId()))
             .collect(Collectors.toList());
     }
@@ -112,7 +113,7 @@ public class CCDCaseApi {
     public List<Claim> getByDefendantEmail(String defendantEmail, String authorisation) {
         User user = userService.getUser(authorisation);
 
-        return asStream(getAllCasesBy(user, ImmutableMap.of()))
+        return asStream(getAllIssuedCasesBy(user, ImmutableMap.of()))
             .filter(claim -> defendantEmail.equals(claim.getDefendantEmail()))
             .collect(Collectors.toList());
     }
@@ -172,12 +173,18 @@ public class CCDCaseApi {
     }
 
     private List<Claim> getAllCasesBy(User user, ImmutableMap<String, String> searchString) {
-        List<CaseDetails> validCases = searchAll(user, searchString)
-            .stream()
-            .filter(caseDetails -> !isCreatedState(caseDetails))
-            .collect(Collectors.toList());
+        return extractClaims(searchAll(user, searchString));
+    }
 
-        return extractClaims(validCases);
+    private List<Claim> getAllIssuedCasesBy(User user, ImmutableMap<String, String> searchString) {
+        List<CaseDetails> issuedCases = asStream(searchAll(user, searchString))
+            .filter(isCaseIssued())
+            .collect(Collectors.toList());
+        return extractClaims(issuedCases);
+    }
+
+    private Predicate<CaseDetails> isCaseIssued() {
+        return caseDetails -> !CREATE.getValue().equals(caseDetails.getState());
     }
 
     private Optional<Claim> getCaseBy(String authorisation, Map<String, String> searchString) {
@@ -394,13 +401,8 @@ public class CCDCaseApi {
     }
 
     private List<Claim> extractClaims(List<CaseDetails> result) {
-        return result
-            .stream()
+        return asStream(result)
             .map(entry -> ccdCaseDataToClaim.extractClaim(entry))
             .collect(Collectors.toList());
-    }
-
-    private boolean isCreatedState(CaseDetails caseDetails) {
-        return CREATE.getValue().equals(caseDetails.getState());
     }
 }
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
@@ -56,6 +56,8 @@ public class CCDCaseApi {
     private static final int MINIMUM_SIZE_TO_CHECK_FOR_MORE_PAGES = 10;
     private static final int MAX_NUM_OF_PAGES_TO_CHECK = 10;
 
+    private Predicate<CaseDetails> isCreatedState = caseDetails -> CREATE.getValue().equals(caseDetails.getState());
+
     @SuppressWarnings("squid:S00107") // All parameters are required here
     public CCDCaseApi(
         CoreCaseDataApi coreCaseDataApi,
@@ -178,13 +180,9 @@ public class CCDCaseApi {
 
     private List<Claim> getAllIssuedCasesBy(User user, ImmutableMap<String, String> searchString) {
         List<CaseDetails> issuedCases = asStream(searchAll(user, searchString))
-            .filter(isCaseIssued())
+            .filter(isCreatedState.negate())
             .collect(Collectors.toList());
         return extractClaims(issuedCases);
-    }
-
-    private Predicate<CaseDetails> isCaseIssued() {
-        return caseDetails -> !CREATE.getValue().equals(caseDetails.getState());
     }
 
     private Optional<Claim> getCaseBy(String authorisation, Map<String, String> searchString) {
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -172,12 +173,18 @@ private void linkToCaseWithoutRevoking(
...
+
+    private List<Claim> getAllIssuedCasesBy(User user, ImmutableMap<String, String> searchString) {
+        List<CaseDetails> issuedCases = asStream(searchAll(user, searchString))
+            .filter(isCaseIssued())
             .collect(Collectors.toList());
+        return extractClaims(issuedCases);
+    }
 
-        return extractClaims(validCases);
+    private Predicate<CaseDetails> isCaseIssued() {
```
karthikkannankainos: `private Predicate<CaseDetails> isCaseIssued = caseDetails -> caseDetails.getState() == CREATE.getValue();` this is the correct way of declaring.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -280,23 +287,26 @@ private boolean isLetterHolderRole(String role) {
         }
 
         User letterHolder = userService.getUser(authorisation);
-        return Optional.of(readCase(letterHolder, letterHolderCases.get(0)));
+        CaseDetails caseDetails = readCase(letterHolder, letterHolderCases.get(0));
+        if (CREATE.getValue().equals(caseDetails.getState())) {
```
karthikkannankainos: Use the predicate created. No need to do if condition again. if you need negate use predicate.not

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -280,23 +287,26 @@ private boolean isLetterHolderRole(String role) {
...
+            throw new DefendantLinkingException("Claim is not in open state, can not link defendant");
```
karthikkannankainos: DefendantLinkingException is mapped to Forbidden errorcode. is that what we want or we can throw badrequestexception? Forbidden means you are not supposed to access that resource.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -172,12 +173,18 @@ private void linkToCaseWithoutRevoking(
...
 
     private List<Claim> getAllCasesBy(User user, ImmutableMap<String, String> searchString) {
-        List<CaseDetails> validCases = searchAll(user, searchString)
-            .stream()
-            .filter(caseDetails -> !isCreatedState(caseDetails))
+        return extractClaims(searchAll(user, searchString));
+    }
+
+    private List<Claim> getAllIssuedCasesBy(User user, ImmutableMap<String, String> searchString) {
+        List<CaseDetails> issuedCases = asStream(searchAll(user, searchString))
```
srineelesh: can you please give this a more clear name. Are you looking to fetch all open claims. In that case maybe rename to 
`getAllOpenCasesBy`

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
```diff
@@ -172,12 +173,18 @@ private void linkToCaseWithoutRevoking(
...
+            .filter(isCaseIssued())
             .collect(Collectors.toList());
+        return extractClaims(issuedCases);
+    }
 
-        return extractClaims(validCases);
+    private Predicate<CaseDetails> isCaseIssued() {
+        return caseDetails -> !CREATE.getValue().equals(caseDetails.getState());
```
srineelesh: If you intend to get only open case I would suggest you to use 
`caseDetails -> !OPEN.getValue().equals(caseDetails.getState());` instead because `!CREATE` can mean closed and settled also

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
sabahirfan: It may include closed or settled cases, `getAllOpenCasesBy` will be incorrect here

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
sabahirfan: done

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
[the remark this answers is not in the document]
sabahirfan: done

sabahirfan: that's exactly what we are looking here anything other than CREATE

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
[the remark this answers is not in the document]
sabahirfan: It's enum description value, should be same all the time, no need for equalsIgnoreCase

sabahirfan: It's simple if condition and not a method reference or lambda expression. So not a good candidate for functional programming here.

karthikkannankainos: Its not to do with Functional Programming. Since you had Predicate already, so we use it. If else, the Predicate you are using becomes unnecessary.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
sabahirfan: Yes, that's exactly what we want to return 403 in that scenario.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
srineelesh: same as above

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
srineelesh: > I don't want to have access to cases that haven't reached the OPEN state yet
Please check the story it says something else..Either have the story corrected or correct the code

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
karthikkannankainos: 403 will be something I try to access someone else case. 
Trying to link myself to open case cannot be 403.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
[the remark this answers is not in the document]
srineelesh: `caseDetails.getState()` -- is a String

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
sabahirfan: The story is also fine. This method is used to fetch cases for dashboard of a defendant which may include cases in all other states i.e. closed or settled.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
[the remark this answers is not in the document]
sabahirfan: `caseDetails.getState()` is coming from the CCD definition and we should not change again. In my opinion, if that is messed up I would rather like it to break our functional tests rather than ignoring case and let them pass.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
sabahirfan: So what are you suggesting here to use then?

karthikkannankainos: 409 Conflict - would be correct. Thats what we use for other business rules validation.
`Indicates that the request could not be processed because of conflict in the current state of the resource` - Conflict explanation

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
karthikkannankainos: Declare once and use multiple times.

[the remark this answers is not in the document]
karthikkannankainos: If we are confident, its always convert the state to Enum and compare with Enum. We created Enum and comparing as String seems wrong.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
karthikkannankainos: If the fetch is for all type of cases. Better to name the method as `getAllCaseCreatedBy` or `getAllCaseBy`.
Issued in our context mean - Created + opened = Issued

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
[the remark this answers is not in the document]
srineelesh: This what the story says **Defendant dashboards don't show cases that haven't been OPEN yet.**
I am not sure how the code and the story are both correct

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
srineelesh: This what the story says **Defendant dashboards don't show cases that haven't been OPEN yet.**
I am not sure how the code and the story are both correct

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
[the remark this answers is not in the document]
srineelesh: > `caseDetails.getState()` is coming from the CCD definition and we should not change again. In my opinion, if that is messed up I would rather like it to break our functional tests rather than ignoring case and let them pass.

I am not saying we should change it .. I am just saying the it doesn't have to be casesensitive..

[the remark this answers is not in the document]
srineelesh: anyways this is just a suggestion.. you can skip this if you don't want to change it.

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseApi.java
sabahirfan: As we discussed, leaving that 403.

PR #1061, opened 2019-07-18
ROC-6050: Update logic for calculating pilot court

sabahirfan:
[ROC-6050](https://tools.hmcts.net/jira/browse/ROC-6050)

### Change description ###
This PR updates the logic for pilot courts

- [ ] Yes
- [x] No

### DIFF
--- a/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/directionsquestionnaire/PilotCourt.java
+++ b/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/directionsquestionnaire/PilotCourt.java
@@ -6,10 +6,11 @@ import java.util.Arrays;
 
 @AllArgsConstructor
 public enum PilotCourt {
-    EDMONTON("Edmonton"),
-    MANCHESTER("Manchester"),
-    BIRMINGHAM("Birmingham"),
-    CLERKENWELL("Clerkenwell & Shoreditch");
+    EDMONTON("edmonton"),
+    MANCHESTER("manchester"),
+    BIRMINGHAM("birmingham"),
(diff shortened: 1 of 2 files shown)


domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/directionsquestionnaire/PilotCourt.java
```diff
@@ -18,7 +19,10 @@ public String getName() {
     }
 
     public static boolean isPilotCourt(String courtName) {
+        if (courtName == null) {
+            return false;
+        }
         return Arrays.stream(PilotCourt.values())
-            .anyMatch(val -> val.name().equalsIgnoreCase(courtName));
+            .anyMatch(pilotCourt -> courtName.toLowerCase().contains(pilotCourt.name));
```
markry-kainos: is contains ok? should it be equals? what if we have something like north birmingham.

sabahirfan: That's exactly what the new logic is. 
```
It is accepted as pilot court if any of the below text is there in the suggested court name

- Manchestor
- Edmonton
- Birmingham
- Clerkenwell 
- Shoreditch

```

markry-kainos: cool 👍 😄

simad: _eyetwitch_

PR #1090, opened 2019-07-31
ROC-6137:Add mediation pdf content for Full defence

KiranVarma02:

https://tools.hmcts.net/jira/browse/ROC-6137

### Change description ###
Add mediation pdf content for Full defence

```
[ ] Yes
[* ] No
```


src/test/java/uk/gov/hmcts/cmc/claimstore/documents/content/DefendantResponseContentProviderTest.java
```diff
@@ -117,4 +117,14 @@ public void shouldProvideResponseTimeline() {
...
             .containsKeys("timelineComment", "events");
     }
+
+    @Test
+    public void shouldProvideMediation() {
+        Map<String, Object> content = provider.createContent(claim);
+
+        assertThat(content)
+            .containsKey("freeMediation")
+            .containsValue("yes")
```
andrewwa-kainos: This 'yes' is not necessarily the `freeMediation`'s value. Consider 
`assertThat(map).containsEntry(key, value)`
 instead

src/main/java/uk/gov/hmcts/cmc/claimstore/documents/content/FullDefenceResponseContentProvider.java
```diff
@@ -46,6 +47,10 @@
             content.put("paymentDeclaration", createContentFor(paymentDeclaration))
         );
 
+        fullDefenceResponse.getFreeMediation().ifPresent(mediation ->
+            content.put("mediation", mediation.equals(YesNoOption.YES))
```
andrewwa-kainos: This puts `true` as the value, but from the other code in this PR it looks like you're expecting `yes` and `no`.
Try `content.put("mediation", mediation.name().toLowerCase())`

karthikkannankainos: Just discussed with @KiranVarma02 . Template expects true or false. The yes or no is determined in template.

src/test/java/uk/gov/hmcts/cmc/claimstore/documents/content/DefendantResponseContentProviderTest.java
KiranVarma02: done

src/main/java/uk/gov/hmcts/cmc/claimstore/documents/content/FullDefenceResponseContentProvider.java
sabahirfan: `mediation == YesNoOption.YES` would be preferred instead of equals here

PR #1123, opened 2019-08-16
Enable DQs in demo only

markry-kainos:
### Change description ###

Enable DQs

```
[ ] Yes
[x] No
```


infrastructure/demo.tfvars
```diff
@@ -8,3 +8,4 @@ ccd_async_enabled = "false"
 dm_url = "http://dm-store-demo.service.core-compute-demo.internal"
 async_event_operations_enabled = "true"
 save_claim_state_enabled = "true"
+directions_questionnaire_enabled = "true"
```
KiranVarma02: what about mediation toggle? is it already true? i can't see it here..

KiranVarma02: @robertparkinson, i forgot if we have mediation toggle!

markry-kainos: no mediation toggles in backend - not relevant, they just drive frontend journeys 🤔

PR #1146, opened 2019-08-28
ROC-6307: Resolve optional fields and document name

dharmendrak:
### Change description ###
ROC-6307: Resolve optional fields and document name

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/callbacks/DrawOrderCallbackHandler.java
```diff
@@ -127,6 +127,7 @@ private CallbackResponse copyDraftToCaseDocument(CallbackParams callbackParams)
         CCDCollectionElement<CCDClaimDocument> claimDocument = CCDCollectionElement.<CCDClaimDocument>builder()
             .value(CCDClaimDocument.builder()
                 .documentLink(draftOrderDoc)
+                .documentName(draftOrderDoc.getDocumentFileName())
```
sabahirfan: can we add size here as well to be consistent with other documents?

dharmendrak: size is not available to add, can look on it later

PR #1171, opened 2019-09-13
ROC-6128: Add channel of case

sabahirfan:
[ROC-6128](https://tools.hmcts.net/jira/browse/ROC-6128)

### Change description ###
Add channel to identify the source of a case.

Definition changes in this PR: https://github.com/hmcts/cmc-ccd-domain/pull/82

- [ ] Yes
- [x] No

### DIFF
--- a/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/CaseMapper.java
+++ b/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/CaseMapper.java
@@ -4,6 +4,8 @@ import org.apache.commons.lang3.EnumUtils;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.cmc.ccd.domain.CCDCase;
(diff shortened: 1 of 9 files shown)


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/CaseMapper.java
```diff
@@ -51,6 +53,11 @@ public CCDCase to(Claim claim) {
             .map(reviewOrderMapper::to)
             .ifPresent(builder::reviewOrder);
 
+        claim.getChannelType()
+            .map(ChannelType::name)
```
simad: as discussed, I'd probably suggest having a dev tech huddle to discuss whether we can start getting rid of a bit of duplication between claim/ccd models. Enums could be a low hanging fruit. But for the scope of this pr is 🆗 of course

sabahirfan: Definitely 👍

PR #1181, opened 2019-09-18
ROC-6491 extend metadata with deadlines and features

andrewwa-kainos:
https://tools.hmcts.net/jira/browse/ROC-6491

### Change description ###
Add missing deadlines and the list of enabled claim features to the case metadata.
Also reordered fields

```
[ ] Yes
[x] No
```


andrewwa-kainos:
> LGTM, hope business doesn't have any objection on seeing some of data to non SC cleared people as well

@dharmendrak that's why we don't include any PII in the metadata

PR #1183, opened 2019-09-19
ROC 6023 mediation pilot app insights and tests

tharacka:

https://tools.hmcts.net/jira/browse/ROC-6023

### Change description ###

Addition of app insights to measure mediation and non-mediation pilot claims that states paid, part admissions or full defence.

- [ ] Yes
- [x] No


src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
```diff
@@ -148,4 +158,16 @@ private boolean shouldFormaliseResponseAcceptance(Response response, ClaimantRes
             && !ResponseUtils.isResponseStatesPaid(response)
             && !ResponseUtils.isResponsePartAdmitPayImmediately(response);
     }
+
+    private boolean isRejectResponseWithMediationForMIReporting(Claim claim, ClaimantResponse claimantResponse) {
```
andrewwa-kainos: Can this just be `isRejectResponseWithMediation` or is that already defined?

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
```diff
@@ -96,6 +96,16 @@ public void save(
 
         ccdEventProducer.createCCDClaimantResponseEvent(claim, claimantResponse, authorization);
         appInsights.trackEvent(getAppInsightsEvent(claimantResponse), "referenceNumber", claim.getReferenceNumber());
+
+        if (isRejectResponseWithMediationForMIReporting(claim, claimantResponse)) {
+            if (claim.getFeatures().contains("mediationPilot")) {
```
andrewwa-kainos: `getFeatures()` can be null for some old cases - add an NPE guard please

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimantResponseService.java
karthikkannankainos: Claim passed is used only to retrieve response. may be we can send the response to this method instead of whole claim

tharacka: It is being used similarly in other methods.

PR #1185, opened 2019-09-20
ROC-6116 Add callback to initiate a payment

simad:

- [x] tests have been updated / new tests has been added (if needed)

https://tools.hmcts.net/jira/browse/ROC-6116

### Change description ###

Adding an endpoint to initiate a citizen payment. This endpoint triggers an event and a ccd callback, in which we calculate fees and create a payment for a citizen.
This endpoint returns either the next url of the payment to frontend, or an error.

```
[ ] Yes
[x] No
```


src/main/resources/application.yml
```diff
@@ -68,6 +68,21 @@ courtfinder:
   api:
     url: "https://courttribunalfinder.service.gov.uk"
 
+fees:
+  api:
+    url: "http://localhost:4182"
```
andrewwa-kainos: These'll need to come from a var, and a `cmc-integration-tests` PR for local envs, right?

simad: yeah should get them from env variables 👍

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ioc/InitiatePaymentResponse.java
```diff
@@ -1,12 +1,16 @@
...
+import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
+import lombok.NoArgsConstructor;
 
 @Data
 @JsonIgnoreProperties(ignoreUnknown = true)
 @Builder
+@AllArgsConstructor
+@NoArgsConstructor
```
KiranVarma02: Just FMI, do we need NoArgsConstructor here? I see that it's used in a test but don't understand purpose.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/InitiatePaymentTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+        verify(coreCaseDataApi)
+            .submitForCitizen(
+                eq(AUTHORISATION_TOKEN),
+                eq(SERVICE_TOKEN),
+                eq(USER_ID),
+                eq(JURISDICTION_ID),
+                eq(CASE_TYPE_ID),
+                eq(IGNORE_WARNING),
+                any()
```
KiranVarma02: Not sure if my understanding is right, should we can verify on the specific object rather than any()?

simad: The assertion is down, we're deserialising the response and checking it is what we expect, so in this case there's no need to be detailed on `verify`. This is an endpoint test - I could've removed the verify all together and the test would still have merit.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ioc/InitiatePaymentResponse.java
simad: You need a default constructor because jackson needs it. @Data and @Value already have some of the needed ones, when you see people being more explicit it's usually because of lombok + jackson issues, like https://github.com/rzwitserloot/lombok/issues/1563. Just don't add extra annotations and if it fails, you'll see why you need one.

KiranVarma02: Thanks for explaining @silvia

src/main/resources/application.yml
```diff
@@ -68,6 +68,21 @@ courtfinder:
...
 
+fees:
+  api:
+    url: ${FEES_URL}
+    service: "civil money claims"
+    jurisdiction1: "civil"
+    jurisdiction2: "county court"
+
+payments:
+  returnUrl: "${FRONTEND_BASE_URL:}/claim/pay/%s/receiver"
```
andrewwa-kainos: Can this explicitly be called a pattern or template or some such?

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/InitiatePaymentTest.java
```diff
@@ -0,0 +1,162 @@
...
+import static uk.gov.hmcts.cmc.ccd.domain.CaseEvent.INITIATE_CLAIM_PAYMENT_CITIZEN;
+import static uk.gov.hmcts.cmc.claimstore.utils.ResourceLoader.successfulCoreCaseDataStoreStartResponse;
+import static uk.gov.hmcts.cmc.claimstore.utils.ResourceLoader.successfulCoreCaseDataStoreSubmitResponse;
+
+@TestPropertySource(
+    properties = {
+        "document_management.url=false",
+        "feature_toggles.ccd_async_enabled=false",
+        "feature_toggles.ccd_enabled=true",
+        "feature_toggles.async_event_operations_enabled=false"
```
sabahirfan: As async operation is in prod now and we are not going to turn that off. We should add tests with this flag set true as default. No need to turn that off for new tests.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/InitiatePaymentTest.java
```diff
@@ -0,0 +1,162 @@
...
+import static org.mockito.Mockito.verify;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+import static uk.gov.hmcts.cmc.ccd.domain.CaseEvent.INITIATE_CLAIM_PAYMENT_CITIZEN;
+import static uk.gov.hmcts.cmc.claimstore.utils.ResourceLoader.successfulCoreCaseDataStoreStartResponse;
+import static uk.gov.hmcts.cmc.claimstore.utils.ResourceLoader.successfulCoreCaseDataStoreSubmitResponse;
+
+@TestPropertySource(
+    properties = {
+        "document_management.url=false",
```
sabahirfan: Is this relevant here?

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/ioc/InitiatePaymentCallbackHandlerTest.java
```diff
@@ -41,23 +90,20 @@ public void shouldStoreIdAndPaymentNextUrlBeforeSubmittingEvent() throws Excepti
...
     private ResultActions makeRequest(String callbackType) throws Exception {
-        CaseDetails caseDetails = CaseDetails.builder()
-            .id(CASE_ID)
-            .data(ImmutableMap.of("data", "existingData"))
-            .build();
-
         CallbackRequest callbackRequest = CallbackRequest.builder()
             .eventId(CaseEvent.INITIATE_CLAIM_PAYMENT_CITIZEN.getValue())
-            .caseDetails(caseDetails)
+            .caseDetails(successfulCoreCaseDataStoreSubmitResponse())
```
sabahirfan: We have started using this `successfulCoreCaseDataStoreSubmitResponse()` extensively, which gives single json from a file. How about using `CaseDetails.CaseDetailsBuilder` to build the relevant CaseDetails object instead? We should populate the only fields CaseDetails would expect in specific scenario (i.e. only the fields we pass in initiate payment callback)

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/InitiatePaymentTest.java
simad: ah right - I think I copy pasted that so there might be another offending test. I'll take a look

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/ioc/InitiatePaymentCallbackHandlerTest.java
simad: sure

src/main/resources/application.yml
simad: 👍

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/InitiatePaymentTest.java
simad: copy paste :D

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/InitiatePaymentTest.java
simad: will set it to true 👍

src/main/java/uk/gov/hmcts/cmc/claimstore/repositories/CCDCaseRepository.java
```diff
@@ -146,8 +144,8 @@ public void requestMoreTimeForResponse(String authorisation, Claim claim, LocalD
     }
 
     @Override
-    public InitiatePaymentResponse initiatePayment(User user, String submitterId, InitiatePaymentRequest data) {
-        return coreCaseDataService.savePayment(user, submitterId, data);
+    public Claim initiatePayment(User user, String submitterId, Claim claim) {
```
sabahirfan: We are not using `submitterId` param, we don't need to pass this

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/CoreCaseDataService.java
```diff
@@ -68,6 +65,7 @@
 
     private static final String CMC_CASE_UPDATE_SUMMARY = "CMC case update";
     private static final String CMC_CASE_CREATE_SUMMARY = "CMC case issue";
+    private static final String CMC_PAYMENT_CREATE_SUMMARY = "CMC payment creatiom";
```
sabahirfan: `creatiom` typo in here

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/callbacks/ioc/PaymentsService.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public PaymentDto createPayment(
+        String authorisation,
+        Claim claim
+    ) {
+
+        logger.info("Calculating interest amount for case {}",
+            claim.getExternalId());
+
+        logger.info("Retrieving fee for case {}",
```
sabahirfan: This log statement seems in wrong place.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/callbacks/ioc/PaymentsService.java
```diff
@@ -0,0 +1,115 @@
...
+            claim.getExternalId());
+
+        BigDecimal amount = claim.getTotalClaimAmount().orElseThrow(IllegalStateException::new);
+        BigDecimal interest = claim.getTotalInterest().orElse(BigDecimal.ZERO);
+
+        BigDecimal amountPlusInterest = amount.add(interest);
+
+        FeeLookupResponseDto feeOutcome = feesClient.lookupFee(
+            "online", "issue", amountPlusInterest
```
sabahirfan: Would be nice to extract these `"online", "issue"` in constants with descriptive names. i.e. `CHANNEL` and `EVENT`

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/callbacks/CallbackParams.java
```diff
@@ -14,6 +14,6 @@
     private Map<Params, Object> params;
 
     public enum Params {
-        BEARER_TOKEN
+        BEARER_TOKEN, SERVICE_AUTH_TOKEN
```
sabahirfan: Can't find any usage for `SERVICE_AUTH_TOKEN`

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/callbacks/ioc/PaymentsService.java
simad: yes, I'll move it a few lines down

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/callbacks/ioc/PaymentsService.java
simad: 👍

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/callbacks/CallbackParams.java
simad: ah yes, sorry, I had added it to verify it was sent in the callback for securing the endpoint :D I'll remove it

PR #1203, opened 2019-10-03
ROC-6581: Restrict IOC legal callback handlers for user in correct role

KiranVarma02:

https://tools.hmcts.net/jira/browse/ROC-6581

### Change description ###
Restrict IOC legal callback handlers for user in correct role


KiranVarma02:
duplicate

PR #1207, opened 2019-10-04
ROC-6581: Role Based Access for Legal

KiranVarma02:

https://tools.hmcts.net/jira/browse/ROC-6581

### Change description ###

Role Based Access for Legal


src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/CreateLegalRepClaimCallbackHandlerTest.java
```diff
@@ -41,7 +41,7 @@ public void setUp() {
         given(referenceNumberRepository.getReferenceNumberForLegal())
             .willReturn(REFERENCE_NO);
 
-        UserDetails userDetails = SampleUserDetails.builder().withRoles("citizen").build();
+        UserDetails userDetails = SampleUserDetails.builder().withRoles("caseworker-cmc-solicitor").build();
```
simad: From https://github.com/hmcts/cmc-claim-store/blob/11ceda09bf0f448445435c97472150dda905f3a5/src/main/java/uk/gov/hmcts/cmc/claimstore/idam/models/UserDetails.java#L64 from that line it seems it should be "solicitor". After a chat with @sabahirfan it seems both "solicitor" and "caseworker-cmc-solicitor" would be fine, but then I question if we should encapsulate this in the `isSolicitor()` method, with an or condition. Otherwise it feels dodgy wanting a role in the callback endpoint, but checking on a slightly different role in the ccd api calls ( cc @dharmendrak)  :)

dharmendrak: from CCD point of view, `solicitor` is not a role it is aware of, instead `caseworker-cmc-solicitor`, it was done in past to make solicitor to use case workers apis as solicitors were not citizens

sabahirfan: We are using case worker endpoints in CCD for solicitor  events, so we have to use `caseworker-cmc-solicitor` role.

PR #1214, opened 2019-10-07
ROC-6121 Create citizen claim endpoint for ioc

simad:

https://tools.hmcts.net/jira/browse/ROC-6121

### Change description ###

Adding endpoint to create a claim, after initiating / resuming a payment

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/ClaimController.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/ClaimController.java
@@ -133,6 +133,17 @@ public class ClaimController {
         return claimService.resumePayment(authorisation, claimData);
     }
 
+    @PutMapping(value = "/{submitterId}/create-citizen-claim", consumes = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    @ApiOperation("Creates a citizen claim")
+    public Claim createClaim(
+        @Valid @NotNull @RequestBody ClaimData claimData,
+        @RequestHeader(HttpHeaders.AUTHORIZATION) String authorisation
+    ) {
+        return claimService.saveCitizenClaim(
+            authorisation,
+            claimData);
+    }
+
     @PutMapping("/defendant/link")
     @ApiOperation("Links defendant to all unlinked letter-holder cases")
     public void linkDefendantToClaim(@RequestHeader(HttpHeaders.AUTHORIZATION) String authorisation) {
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/ClaimController.java
```diff
@@ -133,6 +133,17 @@ public CreatePaymentResponse resumePayment(
         return claimService.resumePayment(authorisation, claimData);
     }
 
+    @PutMapping(value = "/{submitterId}/create-citizen-claim", consumes = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    @ApiOperation("Creates a citizen claim")
+    public Claim createClaim(
+        @Valid @NotNull @RequestBody ClaimData claimData,
+        @RequestHeader(HttpHeaders.AUTHORIZATION) String authorisation
```
sabahirfan: `front-end` passes list of `features` which are enabled on the claim, and we save as part of that claim, we have those missing here.

simad: 👍

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -219,6 +218,27 @@ public CreatePaymentResponse resumePayment(
             .build();
     }
 
+    @LogExecutionTime
+    @Transactional(transactionManager = "transactionManager")
+    public Claim saveCitizenClaim(
+        String authorisation,
+        ClaimData claimData,
```
sabahirfan: We are getting this ClaimData from our endpoint, but we are not storing this in CCD.

simad: yeah, I was imagining the claim data to be the same as the one we already have stored in ccd after the initiate/resume payment. But you're right, let's save it too

sabahirfan: Thanks

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -437,9 +424,26 @@ private Claim buildClaimFrom(
...
+        if (asyncEventOperationEnabled) {
+            eventProducer.createClaimCreatedEvent(
+                savedClaim,
+                user.getUserDetails().getFullName(),
+                authorisation
+            );
+        } else {
+            eventProducer.createClaimIssuedEvent(
+                savedClaim,
+                "to-be-removed",
```
daveoh: This should have been removed ;)

PR #1216, opened 2019-10-08
ROC-6630 Map null payment status

simad:

https://tools.hmcts.net/jira/browse/ROC-6630

### Change description ###

There might be some payment statuses which are not correctly mapped in the enum, and this is creating problems in prob. Reverting to string while we get all of the possible values from devops.

```
[ ] Yes
[x] No
```

### DIFF
--- a/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/PaymentMapper.java
+++ b/ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/PaymentMapper.java
@@ -57,8 +57,12 @@ public class PaymentMapper implements BuilderMapper<CCDCase, Payment, CCDCase.CC
             ccdCase.getPaymentId(),
             moneyMapper.from(ccdCase.getPaymentAmount()),
             ccdCase.getPaymentReference(),
-            ccdCase.getPaymentDateCreated() != null ? ccdCase.getPaymentDateCreated().format(ISO_DATE) : null,
-            PaymentStatus.fromValue(ccdCase.getPaymentStatus()),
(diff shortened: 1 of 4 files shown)


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/PaymentMapper.java
```diff
@@ -57,8 +57,12 @@ public Payment from(CCDCase ccdCase) {
             ccdCase.getPaymentId(),
             moneyMapper.from(ccdCase.getPaymentAmount()),
             ccdCase.getPaymentReference(),
-            ccdCase.getPaymentDateCreated() != null ? ccdCase.getPaymentDateCreated().format(ISO_DATE) : null,
-            PaymentStatus.fromValue(ccdCase.getPaymentStatus()),
+            ccdCase.getPaymentDateCreated() != null
+                ? ccdCase.getPaymentDateCreated().format(ISO_DATE)
+                : null,
+            ccdCase.getPaymentStatus() != null
+                ? PaymentStatus.fromValue(ccdCase.getPaymentStatus())
```
sabahirfan: We have payment status as `Enum` in CMC domain but in CCD side we have this as a `String`. I think we should have this enum on both places.

simad: Yeah, I left it as string because we decided we didn't want to go through a data migration

sabahirfan: Let's discuss that and do that in separately, and merge this as it is to fix the issue in prod.

PR #1232, opened 2019-10-10
ROC-6642: Add intentionToProceedDeadline to claim

daveoh:

https://tools.hmcts.net/jira/browse/ROC-6642

### Change description ###

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/utils/CaseDetailsConverter.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/utils/CaseDetailsConverter.java
@@ -1,12 +1,16 @@
 package uk.gov.hmcts.cmc.claimstore.utils;
 
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.cmc.ccd.domain.CCDCase;
 import uk.gov.hmcts.cmc.ccd.mapper.CaseMapper;
 import uk.gov.hmcts.cmc.claimstore.processors.JsonMapper;
+import uk.gov.hmcts.cmc.claimstore.services.WorkingDayIndicator;
 import uk.gov.hmcts.cmc.domain.models.Claim;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
 
+import java.time.LocalDate;
+import java.time.LocalDateTime;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/cmc/claimstore/utils/CaseDetailsConverter.java
```diff
@@ -15,14 +19,35 @@
 
     private final CaseMapper caseMapper;
     private final JsonMapper jsonMapper;
+    private final int intentionToProceedDeadline;
```
sabahirfan: This definitely is not the right place for this field. I would not expect any change in this utility class for your story.

sabahirfan: @daveoh We should not have any business logic as part of our CCD mappers, so there is no point of injecting WorkingDayIndicator in ccd-adapter. The mappers should be simply mapping to/from. 

I would expect this business logic to be part of the action/trigger to calculate this deadline set this field in `Claim` then mapper will have simple job to map this and save in CCD. When we will read this claim back from CCD, we will simply map it back from CCD to our `Claim` object.

src/main/java/uk/gov/hmcts/cmc/claimstore/utils/CaseDetailsConverter.java
```diff
@@ -41,4 +66,14 @@ private CCDCase extractCCDCase(Map<String, Object> mapData) {
     public Map<String, Object> convertToMap(CCDCase ccdCase) {
         return (Map<String, Object>) jsonMapper.convertValue(ccdCase, Map.class);
     }
+
+    private LocalDate calculateIntentionToProceedDeadline(LocalDateTime respondedDate) {
+        LocalDate deadline = respondedDate.toLocalDate().plusDays(this.intentionToProceedDeadline);
+
+        while (!workingDayIndicator.isWorkingDay(deadline)) {
```
sabahirfan: FYI - To get next working day we have this method to be used `workingDayIndicator.getNextWorkingDay(...)` which has unit tests as well.

src/main/java/uk/gov/hmcts/cmc/claimstore/utils/CaseDetailsConverter.java
daveoh: @sabahirfan This has been discussed, but requires a definition release but we have a time constraint to get this released today. Are you happy to waive this for the short term?

sabahirfan: @daveoh lets discuss that offline

PR #1269, opened 2019-10-29
upgrade java chart version

klawani:
### Change description ###

use chart-java version 2.13.1  to use cpuRequests: '250m' by default
remove redundant config values of  PAY_URL & FEES_URL

```
[ ] Yes
[ x] No
```

### DIFF
--- a/charts/cmc-claim-store/values.aat.template.yaml
+++ b/charts/cmc-claim-store/values.aat.template.yaml
@@ -5,6 +5,8 @@ java:
     PDF_SERVICE_URL: http://cmc-pdf-service-java.software-engineering
     SEND_LETTER_URL: http://rpe-send-letter-service-java.software-engineering
     DOCUMENT_MANAGEMENT_URL: http://dm-store-java.ccd
+    PAY_URL: false
+    FEES_URL: false
     CLAIM_STORE_DB_HOST: preprod-data-lb.moneyclaim.reform.hmcts.net
     CLAIM_STORE_DB_CONNECTION_OPTIONS: '?sslmode=require'
     CMC_DB_USERNAME: 'cmc@cmc-aat'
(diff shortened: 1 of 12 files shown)


charts/cmc-claim-store/values.aat.template.yaml
```diff
@@ -5,8 +5,6 @@ java:
     PDF_SERVICE_URL: http://cmc-pdf-service-java.software-engineering
     SEND_LETTER_URL: http://rpe-send-letter-service-java.software-engineering
     DOCUMENT_MANAGEMENT_URL: http://dm-store-java.ccd
-    PAY_URL: false
-    FEES_URL: false
```
andrewwa-kainos: Please don't remove these.

charts/cmc-claim-store/values.preview.template.yaml
```diff
@@ -35,8 +35,6 @@ java:
     IDAM_API_URL: https://idam-api.aat.platform.hmcts.net
     PDF_SERVICE_URL: http://cmc-pdf-service-aat.service.core-compute-aat.internal
     DOCUMENT_MANAGEMENT_URL: http://dm-store-aat.service.core-compute-aat.internal
-    PAY_URL: false
-    FEES_URL: false
```
andrewwa-kainos: Please don't remove these

charts/cmc-claim-store/values.aat.template.yaml
klawani: don't worry about it ... it's already in values.yaml file which is the default values

charts/cmc-claim-store/values.preview.template.yaml
klawani: same as in aat. It's default in values.yaml file already ...

sabahirfan: These are the default values in `values.yaml`, we should have only those here which we want to override defaults.

PR #1281, opened 2019-11-06
ROC-6821: Make payment optional set in callback

sabahirfan:
[ROC-6821](https://tools.hmcts.net/jira/browse/ROC-6821)

### Change description ###
Make payment and feeAmountInPennies optional as these are being calculated in callback.

- [ ] Yes
- [x] No


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/mapper/ClaimMapper.java
```diff
@@ -92,14 +92,15 @@ public void to(Claim claim, CCDCase.CCDCaseBuilder builder) {
...
         claimData.getTimeline().ifPresent(timeline -> timelineMapper.to(timeline, builder));
         claimData.getEvidence().ifPresent(evidence -> evidenceMapper.to(evidence, builder));
+        claimData.getPayment().ifPresent(payment -> paymentMapper.to(payment, builder));
 
-        paymentMapper.to(claimData.getPayment(), builder);
         interestMapper.to(claimData.getInterest(), builder);
         amountMapper.to(claimData.getAmount(), builder);
         claim.getTotalAmountTillDateOfIssue().map(moneyMapper::to).ifPresent(builder::totalAmount);
+        claimData.getFeeAmountInPennies()
+            .ifPresent(feeAmountInPennies -> builder.feeAmountInPennies(feeAmountInPennies.toString()));
```
andrewwa-kainos: Preference is to keep each chain item as simple as possible:
```
claimData.getFeeAmountInPennies()
    .map(BigInteger::toString)
    .ifPresent(builder::feeAmountInPennies);
```

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
```diff
@@ -55,8 +54,6 @@
     @NotNull
     private final Amount amount;
 
-    @NotNull
-    @Min(0)
```
andrewwa-kainos: Don't we still want the `Min(0)`? I think it won't activate as validation if the value is `null`

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
```diff
@@ -173,8 +170,11 @@ public Boolean isClaimantRepresented() {
     }
 
     @JsonIgnore
-    public BigDecimal getFeesPaidInPounds() {
-        return MonetaryConversions.penniesToPounds(new BigDecimal(feeAmountInPennies));
+    public Optional<BigDecimal> getFeesPaidInPounds() {
+        if (feeAmountInPennies == null) {
+            return Optional.empty();
+        }
+        return Optional.ofNullable(MonetaryConversions.penniesToPounds(new BigDecimal(feeAmountInPennies)));
```
andrewwa-kainos: Can be simplified
```
return Optional.ofNullable(feeAmountInPennies)
    .map(BigDecimal::new)
    .map(MonetaryConversions::penniesToPounds);
```

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
```diff
@@ -196,19 +196,19 @@ public CreatePaymentResponse initiatePayment(
 
         Claim createdClaim = caseRepository.initiatePayment(user, claim);
 
+        Optional<Payment> payment = createdClaim.getClaimData().getPayment();
         return CreatePaymentResponse.builder()
-            .nextUrl(createdClaim.getClaimData().getPayment().getNextUrl())
+            .nextUrl(payment.orElseThrow(IllegalStateException::new).getNextUrl())
```
andrewwa-kainos: May as well throw the exception as soon as we have the optional

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
sabahirfan: done

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ClaimService.java
sabahirfan: `payment` is being initialised in call back now not coming from front-end.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
sabahirfan: That is never going to be invoked as `feeAmountInPennies` will never be passed from front-end after IOC. will always be initiated in callback now. Anyways added that back for current scenario.

domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/ClaimData.java
sabahirfan: done

PR #1287, opened 2019-11-08
ROC-6795 Format statement of truth

Valkrys:
https://tools.hmcts.net/jira/browse/ROC-6795

### Change description ###
Format statement of truth to match prototype

```
[ ] Yes
[ ] No
```


src/main/resources/citizen/templates/document/claimIssueReceipt.html
```diff
@@ -513,52 +514,56 @@
...
           {{ claim.issuedOn }}
         </td>
       </tr>
       {% endif %}
       {% if claim.statementOfTruth is defined and claim.statementOfTruth is not empty %}
-      <tr>
+      <tr class="no-border">
+        <td></td>
+        <td>
+          <strong>Company:</strong>
```
robertparkinson: should this not be in an If statement? From my reading of this company will always be present even if the person signing it is an individual?

src/main/resources/citizen/templates/document/claimIssueReceipt.html
```diff
@@ -141,9 +141,8 @@
   <table>
     <tr>
       <td rowspan="6">
-        <img
```
dharmendrak: is img tag removed?

Valkrys: no, just replaced by the most recent one.

PR #1301, opened 2019-11-19
ROC-6393: Move params out of url as IIS cannot handle + sign

daveoh:
https://tools.hmcts.net/jira/browse/ROC-6393

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/LinkDefendantToClaimViaAuthenticationTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/LinkDefendantToClaimViaAuthenticationTest.java
@@ -2,7 +2,11 @@ package uk.gov.hmcts.cmc.claimstore.controllers;
 
 import org.junit.Test;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
 import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
+import org.springframework.util.LinkedMultiValueMap;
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/controllers/LinkDefendantToClaimViaAuthenticationTest.java
```diff
@@ -42,10 +46,17 @@ public void shouldReturn200HttpStatusAndUpdatedClaimWhenLinkIsSuccessfullySet()
         );
 
         Claim claim = claimStore.saveClaim(SampleClaimData.builder().build());
-        String urlTemplate = String.format(urlFormat, claim.getReferenceNumber(), username, password);
+        String urlTemplate = String.format(urlFormat, claim.getReferenceNumber());
 
-        webClient
-            .perform(put(urlTemplate))
+        MultiValueMap<String, String> params = new LinkedMultiValueMap<>();
```
KiranVarma02: should we use MultiValuedMap instead of MultiValueMap? - 

https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/MultiValueMap.html

Just for my knowledge - do we need MultiValueMap here since we seem to have different values for different keys?

daveoh: The API for the params expects a MultiValueMap. My assumption here is it allows for passing arrays of values for a given key (though why they don't just use an array or collection object here is beyond me).

PR #1307, opened 2019-11-21
ROC-9872: Update quartz cron

daveoh:
https://tools.hmcts.net/jira/browse/ROC-6872

### Change description ###

```
[ ] Yes
[X] No
```


infrastructure/aat.tfvars
```diff
@@ -8,4 +8,4 @@ save_claim_state_enabled = "true"
 milo_csv_schedule = "0 0 8 * * ?"
 directions_questionnaire_enabled = "true"
 doc_assembly_api_url = "http://dg-docassembly-aat.service.core-compute-aat.internal"
-claim_stayed_schedule = "0 0 16 * * 1-5"
+claim_stayed_schedule = "0 0 16 * * 2-6 *"
```
andrewwa-kainos: Can we use MON-FRI instead of 2-6? and is the last one optional?

daveoh: The last one is not optional. It being missing has caused an issue deploying. 
Good call on the MON-FRI.

PR #1308, opened 2019-11-21
disable claim_stayed_schedule

klawani:
## Change description ###

this is to fix misconfiguration of claim_stayed_schedule variable.

```
[ ] Yes
[ ] No
```


infrastructure/variables.tf
```diff
@@ -101,5 +101,5 @@ variable "common_tags" {
 }
 
 variable claim_stayed_schedule  {
-  default = ""
+  default = "-"
```
daveoh: @klawani  - is not a valid cron expression so this will cause errors.

PR #1395, opened 2020-01-21
ROC-7250 Relabel staff notifications. Tidy code.

andrewwa-kainos:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-7250

### Change description ###
Tidy up some of the messy and inconsistent code in the `SupportController`, and reword the staff notification labels to align with the terms used for other support endpoints.

Tools updated accordingly at https://github.com/hmcts/cmc-live-support-tools/pull/10

- [ ] Yes
- [x] No


src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/support/SupportController.java
```diff
@@ -140,22 +141,22 @@ public void resendStaffNotifications(
         String authorisation = user.getAuthorisation();
 
         Claim claim = claimService.getClaimByReferenceAnonymous(referenceNumber)
-                .orElseThrow(() -> new NotFoundException(String.format(CLAIM_DOES_NOT_EXIST, referenceNumber)));
+            .orElseThrow(claimNotFoundException(referenceNumber));
 
         switch (event) {
```
daveoh: `event` should be an enum rather than a string

andrewwa-kainos: Have tried this, but it just ends up with pointless boilerplate. It may be more useful if we were more consistent, but Robotics Support (for example) has a separate endpoint for each type and can't reuse the enum. I'm confident the whole support strategy could use an overhaul, as it's grown rather organically and inconsistently over the last 2 years.

PR #1401, opened 2020-01-23
ROC-7249 Add support to resubmit refer to judge claimant responses

andrewwa-kainos:
### JIRA link  ###
https://tools.hmcts.net/jira/browse/ROC-7249

### Change description ###
Add support to resubmit refer to judge claimant responses

- [ ] Yes
- [x] No


src/main/java/uk/gov/hmcts/cmc/claimstore/events/claimantresponse/ClaimantResponseStaffNotificationHandler.java
```diff
@@ -36,11 +47,17 @@ public void onClaimantResponse(ClaimantResponseEvent event) {
...
             this.statesPaidStaffNotificationService.notifyStaffClaimantResponseStatesPaidSubmittedFor(claim);
         }
-        if (isPartAdmission(response)
-            && claim.getClaimantResponse().orElseThrow(IllegalArgumentException::new).getType()
-            == ClaimantResponseType.REJECTION) {
+        ClaimantResponse claimantResponse = claim.getClaimantResponse().orElseThrow(IllegalArgumentException::new);
+        if (isPartAdmission(response) && claimantResponse.getType() == ClaimantResponseType.REJECTION) {
             claimantRejectionStaffNotificationService.notifyStaffClaimantRejectPartAdmission(claim);
         }
+        if (isReferredToJudge(claimantResponse)) {
```
daveoh: The nested if can be avoided with

```suggestion
        if (!isReferredToJudge(claimantResponse)) {
            return;
        }
        
         if (isCompanyOrOrganisation(response.getDefendant())) {
         ....
```

andrewwa-kainos: I'd prefer not to explicitly `return` here, in case we end up adding new claimant-response conditions under which a staff notification should be sent.

PR #1429, opened 2020-02-06
ROC-7498 Mediation successful template missing claim reference in fai…

dharmendrak:

https://tools.hmcts.net/jira/browse/ROC-7498

### Change description ###

ROC-7498 Mediation successful template missing claim reference in fai…

- [ ] Yes
- [ ] No


src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/callbacks/mediation/MediationSuccessfulCallbackHandler.java
```diff
@@ -96,8 +96,7 @@ private void notifyClaimant(Claim claim) {
             claim.getSubmitterEmail(),
             notificationsProperties.getTemplates().getEmail().getClaimantMediationSuccess(),
             aggregateParams(claim),
-            MediationSuccessful
-                .referenceForClaimant(claim.getReferenceNumber(), "claimant")
+            MediationSuccessful.referenceForClaimant(claim.getReferenceNumber(), "claimant")
```
andrewwa-kainos: This will be:
`customDimensions = {"referenceNumber":"to-defendant-mediation-successful-012MC345"}`

Probably originally we only had one type of notification failure and just use the claim reference number. If we want to use the same thing then I suggest a change to the `customDimensions` we report to include both the claim reference number and the notification type. As it is, this doesn't make sense as read.

dharmendrak: I agree @andrewwa-kainos , however that change would be bit bigger, best would be to have a tech debt for that and complete it.

PR #1432, opened 2020-02-06
Remove the deprecated `enableDockerBuild()` configuration

andrewwa-kainos:
### JIRA link ###
N/A

### Change description ###
Remove the `enableDockerBuild()` configuration from Jenkins_CNP

- [ ] Yes
- [x] No

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -63,6 +63,5 @@ withPipeline("java", product, component) {
 
   loadVaultSecrets(secrets)
   enableSlackNotifications(notificationsChannel)
-  enableDockerBuild()
   installCharts()
 }


Jenkinsfile_CNP
```diff
@@ -63,6 +63,5 @@ withPipeline("java", product, component) {
 
   loadVaultSecrets(secrets)
   enableSlackNotifications(notificationsChannel)
-  enableDockerBuild()
```
KiranVarma02: what did it do? do we need some new configuration?

andrewwa-kainos: The message from Jenkins is:
```
enableDockerBuild() is deprecated, a Dockerfile has been mandatory since 17/12/2019, please remove this option from your Jenkinsfile This configuration will stop working by 18/02/2020 00:00 AM ( in 11 days )
```
So it looks like it was doing something that is now the default anyway.

PR #1449, opened 2020-02-11
ROC-7508-Flyway fails on migrate unable to find flyway_schema_history table

karthikkannankainos:

https://tools.hmcts.net/jira/browse/ROC-7508

### Change description ###

1. To solve it a solution on flyway doc to set the
baselineOnMigrate to true.

### DIFF
--- a/src/main/java/uk/gov/hmcts/cmc/claimstore/config/CmcDBConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/cmc/claimstore/config/CmcDBConfiguration.java
@@ -46,6 +46,7 @@ public class CmcDBConfiguration {
     private void migrateFlyway(DataSource dataSource) {
         Flyway.configure()
             .dataSource(dataSource)
+            .baselineOnMigrate(true)
             .locations("db/migration")
             .load()
             .migrate();


src/main/java/uk/gov/hmcts/cmc/claimstore/config/CmcDBConfiguration.java
```diff
@@ -46,6 +46,7 @@ public TransactionAwareDataSourceProxy transactionAwareDataSourceProxy(
     private void migrateFlyway(DataSource dataSource) {
         Flyway.configure()
             .dataSource(dataSource)
+            .baselineOnMigrate(true)
```
andrewwa-kainos: Is there anything in place - code or practice - to mitigate the risk described in the documentation?

> Be careful when enabling this as it removes the safety net that ensures Flyway does not migrate the wrong database in case of a configuration mistake!

karthikkannankainos: I think this might for the multiple database. We have currently only single database CMC.

sabahirfan: I don't think we need to enable this here.

PR #1452, opened 2020-02-12
ROC-6863: Configure per pilot courts

daveoh:
https://tools.hmcts.net/jira/browse/ROC-6863

### Change description ###
Configure different courts for different pilots

- [ ] Yes
- [X] No


src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/callbacks/legaladvisor/OrderCreator.java
```diff
@@ -203,4 +205,18 @@ private boolean hasRequestedForPermissionWithProvidedEvidence(CCDDirectionsQuest
...
+        CaseEvent caseEvent = CaseEvent.fromValue(callbackParams.getRequest().getEventId());
+        switch (caseEvent) {
+            case GENERATE_ORDER:
+                return Pilot.LA;
+            case ACTION_REVIEW_COMMENTS:
+                return Pilot.LA;
+            case DRAW_JUDGES_ORDER:
+                return Pilot.JDDO;
+            default:
+                throw new IllegalArgumentException(String.format("No pilot defined for event: %s", caseEvent));
```
andrewwa-kainos: Concatenation would be cleaner and more efficient

src/main/java/uk/gov/hmcts/cmc/claimstore/services/pilotcourt/PilotCourtService.java
```diff
@@ -64,14 +66,24 @@ public void init() {
         return pilotCourts.keySet();
     }
 
+    public Set<String> getPilotCourtIds(Pilot pilot) {
```
andrewwa-kainos: Isn't this making assumptions about how the consumer wants to handle the court data? I would expect the Pilot Court Service to return collections of Courts and let the consumer continue processing as they wish. (Also, since it also has methods returning Hearing Courts, wouldn't this class be better named as just a Court Service?)

src/main/java/uk/gov/hmcts/cmc/claimstore/services/pilotcourt/PilotCourtService.java
```diff
@@ -95,25 +116,43 @@ public boolean isPilotCourt(String courtName) {
 
         Map<String, PilotCourt> pilotCourts = new HashMap<>();
         while (iterator.hasNext()) {
-            CSVRecord next = iterator.next();
+            CSVRecord csvRecord = iterator.next();
+
+            assert (csvRecord.size() == PilotCourtCSVHeader.values().length) : String.format("Pilot court "
+                + "configuration for %s is expected to have %d values",
+                csvRecord.toString(),
+                PilotCourtCSVHeader.values().length);
```
andrewwa-kainos: I doubt assertions are enabled in prod

daveoh: If they aren't that is a bad practice, unless the assertions cause a significant performance hit. However it shouldn't matter as the configuration it asserts is something that is added in development, not on an environmental level.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/ccd/callbacks/legaladvisor/OrderCreator.java
```diff
@@ -216,7 +220,7 @@ private Pilot getPilot(CallbackParams callbackParams) {
             case DRAW_JUDGES_ORDER:
                 return Pilot.JDDO;
             default:
-                throw new IllegalArgumentException(String.format("No pilot defined for event: %s", caseEvent));
+                throw new IllegalArgumentException(String.format("No pilot defined for event: " + caseEvent));
```
andrewwa-kainos: Now `String.format()` isn't needed at all :)

src/main/java/uk/gov/hmcts/cmc/claimstore/services/pilotcourt/PilotCourtService.java
andrewwa-kainos: Disagree re: bad practice, but either way this would be the only place in claim-store where it'd be used - no real value in deviating.

PR #1459, opened 2020-02-13
ROC-7511 release courts on date

daveoh:
https://tools.hmcts.net/jira/browse/ROC-7511

### Change description ###
Courts to be made available for claims created after a certain date

- [ ] Yes
- [X] No


src/main/java/uk/gov/hmcts/cmc/claimstore/services/pilotcourt/PilotCourtService.java
```diff
@@ -144,15 +152,38 @@ private boolean checkPilotCourt(PilotCourt pilotCourt, String courtName, Pilot p
...
+        .map(pilotCourtCSVHeader -> {
+            String pilotConfig = csvRecord.get(pilotCourtCSVHeader.ordinal());
+
+            if (!isValidPilotDate(pilotConfig)) {
+                return null;
+            }
+
+            return new Object[]{
+                pilotCourtCSVHeader.getPilot().get(),
+                LocalDate.parse(pilotConfig, ISO_LOCAL_DATE).atTime(PILOT_COURT_GO_LIVE_TIME, 0)
```
andrewwa-kainos: Eek. I think if we have to resort to arbitrary structures like this it might be clearer using pre-8 language
```java
        final Map<Pilot, LocalDateTime> results = new HashMap<>();
        for (PilotCourtCSVHeader header : PilotCourtCSVHeader.values()) {
            Optional<Pilot> pilot = header.getPilot();
            if (!pilot.isPresent()) {
                continue;
            }
            String pilotConfig = csvRecord.get(header.ordinal());
            if (!isValidPilotDate(pilotConfig)) {
                continue;
            }
            results.put(
                pilot.get(),
                LocalDate.parse(pilotConfig, ISO_LOCAL_DATE).atTime(PILOT_COURT_GO_LIVE_TIME, 0)
            );
        }
        return results;
```

src/main/java/uk/gov/hmcts/cmc/claimstore/services/pilotcourt/PilotCourtService.java
```diff
@@ -144,15 +152,38 @@ private boolean checkPilotCourt(PilotCourt pilotCourt, String courtName, Pilot p
...
+            };
+
+        })
+        .filter(Objects::nonNull)
+        .collect(Collectors.toMap(e -> (Pilot)e[0], e -> (LocalDateTime) e[1]));
+    }
+
+    // Valid configuration for a pilot trigger date is either a date in IOS format (yyyy-MM-dd) or false
+    private boolean isValidPilotDate(String pilotConfig) {
```
andrewwa-kainos: This is confusing. The comment says valid = ISO date* or 'false', which suggests 'false' is valid, but 'isValid...' returns false for 'false'...

\* note, also a typo, and should specify ISO-8601 as there are other ISO date formats... buuut probably readers will assume correctly. Also, if a method needs commenting, better as javadoc.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/pilotcourt/PilotCourtService.java
```diff
@@ -144,15 +152,38 @@ private boolean checkPilotCourt(PilotCourt pilotCourt, String courtName, Pilot p
...
+        try {
+            LocalDate.parse(pilotConfig, ISO_LOCAL_DATE);
+        } catch (DateTimeParseException e) {
+            if (!pilotConfig.equals("false")) {
+                throw new IllegalArgumentException(String.format("Expecting configuration to either be a valid date or "
+                    + "'false' but got %s", pilotConfig));
```
andrewwa-kainos: Could we possibly detect malformatting earlier in the PostConstruct process? Regex perhaps. Then this part of the code could be cleaner.

src/main/java/uk/gov/hmcts/cmc/claimstore/services/pilotcourt/PilotCourtService.java
daveoh: The intention here is to make the configuration clear. The comment is to make the code clear.

daveoh: I think changing the method name will make things clearer

src/main/java/uk/gov/hmcts/cmc/claimstore/services/pilotcourt/PilotCourtService.java
daveoh: Regex is not clean!

daveoh: Personally I think reading and validating in a single pass is better.

PR #1520, opened 2020-03-01
ROC-7380 - decommission paid in full staff emails

tharacka:
# JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/ROC-7380

### Change description ###
decommision staff emails for paid in full

- [ ] Yes
- [x] No


src/main/java/uk/gov/hmcts/cmc/claimstore/services/staff/PaidInFullStaffNotificationService.java
```diff
@@ -15,6 +16,7 @@
 import java.util.Map;
 
 @Service
+@ConditionalOnProperty("feature_toggles.staff_emails_enabled")
```
akritis22: could have done on handler level as the service is used only on the handler. The new instance of handler won't be created then in first place.

PR #1538, opened 2020-03-05
ROC-7628: Revert the ROC-7612 change introduced in system due send letter client version upgrade

dharmendrak:

https://tools.hmcts.net/jira/browse/ROC-7626

### Change description ###

Revert the ROC-7612 change introduced in system due send letter client version upgrade

- [ ] Yes
- [ ] No


src/main/java/uk/gov/hmcts/cmc/claimstore/documents/BulkPrintService.java
```diff
@@ -146,7 +147,8 @@ public void printPdf(Claim claim, List<Printable> documents) {
             return document.template.getBytes();
         }
 
-        return pdfServiceClient.generateFromHtml(document.template.getBytes(), document.values);
+        byte[] html = pdfServiceClient.generateFromHtml(document.template.getBytes(), document.values);
```
timja: have you tested this change anywhere?

dharmendrak: I have reverted back to old code as it was removed by mistake

src/main/java/uk/gov/hmcts/cmc/claimstore/documents/BulkPrintService.java
```diff
@@ -146,7 +147,8 @@ public void printPdf(Claim claim, List<Printable> documents) {
...
+        return Base64.getEncoder().encodeToString(html).getBytes();
```
njrich28:
```suggestion
        return Base64.getEncoder().encode(html);
```

No need to encode to `String` and back to `byte[]` again. Also, `getBytes` uses the platform default character encoding but `encodeToString` always uses ISO-8859-1 so you may run into encoding issues.

dharmendrak: yes it is failing , so looking into it

src/main/java/uk/gov/hmcts/cmc/claimstore/documents/BulkPrintService.java
pragnesh-patel: We need tests

PR #1598, opened 2020-03-26
[WIP] ROC-7745: add spring boot security to implement openid

sabahirfan:
[ROC-774](https://tools.hmcts.net/jira/browse/ROC-7745)

### Change description ###
Add spring boot security to claim-store. This will use the default spring boot handlers for all of that stuff related to OIDC.

- [ ] Yes
- [x] No


build.gradle
```diff
@@ -282,6 +292,7 @@ task apiTest(type: Test) {
   description = 'Runs the api tests'
   setTestClassesDirs(sourceSets.integrationTest.output.classesDirs)
   setClasspath(sourceSets.integrationTest.runtimeClasspath)
+  testLogging.showStandardStreams = true
```
dharmendrak, line 295: please remove it

sabahirfan: 👍

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/BaseMockSpringTest.java
```diff
@@ -136,6 +168,48 @@
...
+            .collect(Collectors.toCollection(ArrayList::new));
+
+        when(authenticationMock.getAuthorities()).thenAnswer(invocationOnMock -> authorityCollection);
+
+    }
+
+    @NotNull
+    private Jwt getJwt() {
+        return Jwt.withTokenValue(BEARER_TOKEN)
+            .claim("exp", Instant.ofEpochSecond(1585763216))
```
dharmendrak, line 202: this timestamp can expire in future and test will start failing, so we need to, take current date and add some time in it

sabahirfan: I think only the format is being validate as we are mocking getuserinfo calls.

PR #1609, opened 2020-03-31
ROC-7355 ROC-7797 Integration tests for the CCJ component

andrewwa-kainos:
Parent: https://tools.hmcts.net/jira/browse/ROC-7355
Sub-task: https://tools.hmcts.net/jira/browse/ROC-7797

### Change description ###
Add integration tests for the CCJ component

- [ ] Yes
- [x] No


src/main/java/uk/gov/hmcts/cmc/claimstore/services/CountyCourtJudgmentService.java
```diff
@@ -66,11 +63,11 @@ public Claim save(
 
         authorisationService.assertIsSubmitterOnClaim(claim, userDetails.getId());
 
-        countyCourtJudgmentRule.assertCountyCourtJudgementCanBeRequested(claim, countyCourtJudgment.getCcjType());
+        countyCourtJudgmentRule.assertCountyCourtJudgmentCanBeRequested(claim, countyCourtJudgment.getCcjType());
 
-        Optional<Response> response = claim.getResponse();
-        if (claim.getState().equals(ClaimState.STAYED)
-            && response.filter(ResponseUtils::isAdmissionResponse).isPresent()) {
+        if (claim.getState() == ClaimState.STAYED
```
justin-r-oliver, line 68: Why change? Its generally better to use equals() where the LHS could be null.

andrewwa-kainos:
```
definitelyPresent.equals(possiblyNull)
```
is preferred over
```
possiblyNull.equals(definitelyPresent)
```
but with `enum`s, we can just use `==`, which is null-safe. Remember: enums are explicitly singletons.

PR #1633, opened 2020-04-07
ROC-7696 Add paymentDate and paymentMethod to defence response pdf

justin-r-oliver:
**JIRA link**
https://tools.hmcts.net/jira/browse/ROC-7696

**Change description**
1. Added paymentDate and paymentMethod to defence response PDF for states paid
2. Formatted document template to improve readability e.g. comments, indenting
3. Automated testing of document generation using a HTML document comparison approach


src/main/resources/citizen/templates/document/defendantResponseReceipt.html
```diff
@@ -236,17 +236,21 @@
       <td colspan="2"><p><strong>Claim number: </strong>{{ claim.referenceNumber }}</p></td>
     </tr>
     <tr>
-      <td colspan="2"><p><strong>Issued on: </strong> {{ claim.issuedOn }}</p></td>
+      <td colspan="2"><p><strong>Issued on: </strong>
+        <span id="timestamp">{{ claim.issuedOn }}</span>
```
andrewwa-kainos, line 240: `id`s should be unique in a document. Consider using a `class` instead.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/documents/DefendantResponseReceiptServiceIntegrationTest.java
```diff
@@ -0,0 +1,98 @@
...
+        byte[] actualHtmlBytes = defendantResponseReceiptService.createHtml(claim);
+
+        String actualHtml = documentComparisonHelper.replaceTimestamp("timestamp",
+            new String(actualHtmlBytes));
+
+        writeFile("build/tmp/actual.html", actualHtml); // Useful file for debugging test
+
+        String expectedHtml = readFile("src/integrationTest/resources/documents/expectedDefendantResponseReceipt.html");
+
+        assertXMLEqual(actualHtml, expectedHtml);
```
andrewwa-kainos, line 81: I'm not sure this is the right approach; we'll have to mirror any changes to the template in the expected document. The test shouldn't fail just because we changed a font size or switched from `<ol>` to `<ul>`.

Instead, I suggest searching the document for appropriate text according to the input. In this case, you created a full defence, already paid; so to verify it we'd look for "I have paid what I believe I owe". Then you could add a couple more scenarios covering the high-level variations, and we won't need the xmlAssert dependency.

Let me know what you think!

.travis.yml
```diff
@@ -17,7 +17,7 @@ cache:
   - $HOME/.gradle/caches/
   - $HOME/.gradle/wrapper/
 script:
-  - ./gradlew check
```
andrewwa-kainos, line 20: Don't forget to revert this, please

justin-r-oliver: Reverted

src/main/resources/citizen/templates/document/defendantResponseReceipt.html
justin-r-oliver: Next changes uses data-type instead of id

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/documents/DefendantResponseReceiptServiceIntegrationTest.java
justin-r-oliver: I could use a simple text equals comparison instead of xmlAssert. However, the advantages of xmlAssert are that it will validate that the document is well formed, and it will identify the location of any differences using XPath. 

You are right that doing a HTML document comparison will mean that some trivial changes will cause test failures. But it also means that it will detect formatting regressions that would not be found by doing specific text comparisons. It is also is an easy way to get full coverage of the document content.

src/integrationTest/java/uk/gov/hmcts/cmc/claimstore/documents/DefendantResponseReceiptServiceIT.java
```diff
@@ -0,0 +1,94 @@
...
+            )
+            .withRespondedAt(LocalDateTime.now())
+            .build();
+
+        byte[] actualHtmlBytes = defendantResponseReceiptService.createHtml(claim);
+
+        String actualHtml = replaceTimestampsWithFixedValues(new String(actualHtmlBytes));
+
+        // Useful for debugging test ie diff comparison of actual with expected HTML file
+        writeFile("build/tmp/actual.html", actualHtml);
```
andrewwa-kainos, line 64: Not sure this should be kept in the code; but if it is, it can be refactored to be guaranteed platform-independent.

Change the signature to `writeFile(String actualHtml, String firstPath, String... morePaths)`
and build the path with `Paths.get(firstPath, morePaths)` ([javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Paths.html#get(java.lang.String,java.lang.String...)))
and call the method as `writeFile(actualHtml, "build", "tmp", "actual.html")`

You can also read the contents of a file more simply like this: `Files.readString(path)` ([javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#readString(java.nio.file.Path)))

justin-r-oliver: Simplified IO in next commit

gradle.properties
```diff
@@ -0,0 +1,2 @@
+
+org.gradle.jvmargs=-XX:UseAVX=0
```
andrewwa-kainos, line 2: This is not a requirement of the project - please revert.
(You can configure such JVM options locally as necessary.)

sabahirfan: please remove this.

justin-r-oliver: Added comments to justify the JVM setting

sabahirfan: Thanks

justin-r-oliver:
This PR is still required.

PR #1638, opened 2020-04-07
ROC-7242 refer to judge callback

akritis22:
https://tools.hmcts.net/jira/browse/ROC-7242

### Change description ###
A new callback has been added which changed "assignedTo" to OCMC boxwork so that the judge can filter based on assign to and see the case.

- [ ] Yes
- [x] No


ccd-adapter/src/main/java/uk/gov/hmcts/cmc/ccd/domain/CCDLaList.java
```diff
@@ -0,0 +1,34 @@
+package uk.gov.hmcts.cmc.ccd.domain;
+
+public enum CCDLaList {
+
+    REFER_TO_JUDGE("OCMC boxwork"),
+    REFER_TO_COURT_STAFF("Any courtstaff with direction"),
```
daveoh, line 6: Court Staff needs to be 2 words

akritis22: Actually removing **REFER_TO_COURT_STAFF("Any courtstaff with direction"),** this is not relevant to this story.

PR #1647, opened 2020-04-15
Update access for citizen for all new events

dharmendrak:


src/main/java/uk/gov/hmcts/cmc/claimstore/services/DirectionsQuestionnaireService.java
```diff
@@ -45,7 +46,7 @@ public CaseEvent prepareCaseEvent(ResponseRejection responseRejection, Claim cla
             return REFERRED_TO_MEDIATION;
         }
 
-        if (!isOnlineDQ(claim)) {
+        if (!isOnlineDQ(claim) && ClaimantResponseHelper.hasNotOptedForMediation(responseRejection)) {
```
daveoh, line 49: Is this exta clause needed needed? The if statement above this is `if (isOptedForMediation(responseRejection))`

PR #1648, opened 2020-04-16
Resolve new definition based failures

dharmendrak:


src/main/java/uk/gov/hmcts/cmc/claimstore/services/DirectionsQuestionnaireService.java
```diff
@@ -45,7 +46,7 @@ public CaseEvent prepareCaseEvent(ResponseRejection responseRejection, Claim cla
             return REFERRED_TO_MEDIATION;
         }
 
-        if (!isOnlineDQ(claim)) {
+        if (!isOnlineDQ(claim) && ClaimantResponseHelper.hasNotOptedForMediation(responseRejection)) {
```
daveoh, line 49: Is this clause needed. The previous if already checks that mediation has not been opted for.

dharmendrak: not required

PR #1766, opened 2020-07-03
Feature/roc 7944 reconsideration deadline change

shashisk97:

https://tools.hmcts.net/jira/browse/ROC-7944

### Change description ###

Reconsideration: Either party will have 7 days to request a reconsideration instead of a 19 day deadline 

- [ ] Yes
- [x] No


src/main/resources/application.yml
```diff
@@ -139,6 +139,10 @@ dateCalculations:
   waitingTransferDeadlineInDays: 19
   claimantCcjReminderDays: 10
 
+directionDeadline:
+  changeDate: "2020-07-27T11:00:00"
```
singhtanveer, line 143: Can you just once confirm if this should be hardcoded here or taken from environment variable?

singhtanveer, line 143: Can you just once confirm if this should be hardcoded here or taken from environment variable?

shashisk97: This config is added to maintain backward compatibality and it is not going to change so can be there as it is.

PR #2712, opened 2023-02-28
CIV-7300 addition of hearing reference sequence

drummondjm:
Addition of hearing reference sequence required for generating hearing reference number 

As part of https://tools.hmcts.net/jira/browse/CIV-7300

### Change description ###

- [ ] Yes
- [ x ] No


filippo-possenti-version1 (approved):

filippo-possenti-version1:
A bit of context on this commit:
this was rendered necessary as Demo appears to be using the cmc-claim-store database. While the script was prepared last year, it was committed in civil-service and as a result never got deployed in demo. This commit copies the script file we had in civil-service within the cmc-claim-store repository so that we'll then be able to rebuild demo appropriately and correctly verify one of our features.

sabahirfan (changes requested):
If this is for civil-service repo and new service, I would recommend not add here in claim-store repo. As this is deprecated service and going to be decommissioned any time soon (was supposed to be decommissioned mid last year). 

So let's not add any more dependencies on this one for new service.

PR #2819, opened 2023-12-05
CIV-10543 pointing to v15

asthamalviya:
For release final PR


charts/cmc-claim-store/values.aat.template.yaml
```diff
@@ -7,5 +7,5 @@ java:
     CLAIM_STAYED_SCHEDULE: "0 0 16 ? * MON-FRI *"
     SCHEDULE_TRANSFER_CLAIMS: '0 0 12 1 * ?'
     OIDC_ISSUER: https://forgerock-am.service.core-compute-idam-aat2.internal:8443/openam/oauth2/realms/root/realms/hmcts
-    CMC_DB_USERNAME: cmc@cmc-db-v11-aat
-    CMC_DB_HOST: cmc-db-v11-aat.postgres.database.azure.com
+    CMC_DB_USERNAME: cmc@cmc-db-v15-aat
```
dharmendrak, line 10: we don't need here

charts/cmc-claim-store/values.aat.template.yaml
```diff
@@ -7,5 +7,5 @@ java:
...
+    CMC_DB_HOST: cmc-db-v15-aat.postgres.database.azure.com
```
dharmendrak, line 11: same as above

dharmendrak (changes requested):

charts/cmc-claim-store/values.aat.template.yaml
asthamalviya: Done as suggested

